PR #8, opened 2025-01-18
Making it drive

iamawesomecat:
Assigned controls, taken from the beta bot, but only for the drivebase. removed some unnecessary fluff.

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -47,6 +47,10 @@ public class Controls {
   }
 
   private void configureDrivebaseBindings() {
+    if (s.drivebaseSubsystem == null) {
+      // Stop running this method
+      return;
+    }
     // Note that X is defined as forward according to WPILib convention,
     // and Y is defined as to the left according to WPILib convention.
     s.drivebaseSubsystem.setDefaultCommand(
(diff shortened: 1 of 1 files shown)


src/main/java/frc/robot/Controls.java
```diff
@@ -15,9 +20,61 @@ public class Controls {
...
+  private final Telemetry logger = new Telemetry(MaxSpeed);
+
   public Controls(Subsystems s) {
     driverController = new CommandXboxController(DRIVER_CONTROLLER_PORT);
     operatorController = new CommandXboxController(OPERATOR_CONTROLLER_PORT);
     this.s = s;
+    configureDrivebaseBindings();
+  }
+
+  private void configureDrivebaseBindings() {
```
KangarooKoala, line 49:
```suggestion
  private void configureDrivebaseBindings() {
    if (s.drivebaseSubsystem == null) {
      // Stop running this method
      return;
    }
```
We could also do the null check in the constructor or check the enabled constant, but this is the most direct way to not configure drivebase bindings when we don't have a drivebase subsystem object.

KangarooKoala (changes requested):

src/main/java/frc/robot/Controls.java
iamawesomecat: Added. Commit will be pushed momentarily

KangarooKoala (approved):

PR #9, opened 2025-01-18
Elevator Subsystem 1-18 creation

braelynandthefrogs:
added test elevator code. should zero out any current values before testing

### DIFF
--- a/src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
@@ -103,7 +103,7 @@ public class ElevatorSubsystem extends SubsystemBase {
     m_motor.setVoltage(pidOutput + feedforwardOutput);
   }
 
-  public void motorConfigs(){
+  public void motorConfigs() {
     var talonFXConfigurator = m_motor.getConfigurator();
     var talonFXConfigurator2 = m_motor2.getConfigurator();
     var currentLimits = new CurrentLimitsConfigs();
--- a/src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
@@ -1,0 +1,3 @@
+package frc.robot.subsystems;
+
+public class ElevatorSubsystem {}
--- a/src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
@@ -10,8 +10,6 @@ import edu.wpi.first.math.controller.ElevatorFeedforward;
 import edu.wpi.first.math.controller.ProfiledPIDController;
 import edu.wpi.first.math.trajectory.TrapezoidProfile;
 import edu.wpi.first.units.Units;
-import edu.wpi.first.units.measure.MutAngle;
-import edu.wpi.first.units.measure.MutAngularVelocity;
 import edu.wpi.first.units.measure.MutVoltage;
 import edu.wpi.first.units.measure.Voltage;
 import edu.wpi.first.wpilibj.RobotController;
@@ -130,12 +128,14 @@ public class ElevatorSubsystem extends SubsystemBase {
 
   public Command goUp() {
     return startEnd(
-        () -> m_motor.set(UP_VOLTAGE), () -> m_motor.set(HOLD_VOLTAGE)); // test values from dev model bot
+        () -> m_motor.set(UP_VOLTAGE),
+        () -> m_motor.set(HOLD_VOLTAGE)); // test values from dev model bot
   }
 
   public Command goDown() {
     return startEnd(
-        () -> m_motor.set(DOWN_VOLTAGE), () -> m_motor.set(HOLD_VOLTAGE)); // test values from dev model bot
+        () -> m_motor.set(DOWN_VOLTAGE),
+        () -> m_motor.set(HOLD_VOLTAGE)); // test values from dev model bot
   }
 
   /** Stop the control loop and motor output. */
--- a/src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
@@ -29,8 +29,11 @@ public class ElevatorSubsystem extends SubsystemBase {
   private final double ELEVATOR_KV = 0;
   private final double ELEVATOR_KG = 0;
   private final double ELEVATOR_KA = 0;
-  private final double reverseSoftLimit = -67;
-  private final double forwardSoftLimit = -1;
+  private final double REVERSE_SOFT_LIMIT = -67;
+  private final double FORWARD_SOFT_LIMIT = -1;
+  private final double UP_VOLTAGE = -0.25;
+  private final double DOWN_VOLTAGE = 0.04;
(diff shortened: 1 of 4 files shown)


src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,144 @@
...
+  private final double ELEVATOR_KD = 0;
+  private final double ELEVATOR_KS = 0;
+  private final double ELEVATOR_KV = 0;
+  private final double ELEVATOR_KG = 0;
+  private final double ELEVATOR_KA = 0;
+  private final double reverseSoftLimit = -67;
+  private final double forwardSoftLimit = -1;
+
+  // Standard classes for controlling our elevator
+  private final ProfiledPIDController m_controller =
```
jamesdooley4, line 36: Since this is a CTRE-based motor controller (the TalonFX is made by CTRE), using https://v6.docs.ctr-electronics.com/en/latest/docs/api-reference/device-specific/talonfx/motion-magic.html#motion-magic-expo should get better results since the PID & motion profile run on the motor itself at a much higher rate than the Rio can run code & update the motor.

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,144 @@
...
+      new ProfiledPIDController(
+          ELEVATOR_KP, ELEVATOR_KI, ELEVATOR_KD, new TrapezoidProfile.Constraints(2.45, 2.45));
+  ElevatorFeedforward m_feedforward =
```
markpete, line 39: Does this ever get reset or should it be private final as well?

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,144 @@
...
+      new ElevatorFeedforward(ELEVATOR_KS, ELEVATOR_KG, ELEVATOR_KV, ELEVATOR_KA);
+
```
markpete, line 41: Can either m_controller or m_feedforward ever return null or throw an error?  This would be a good place to check it and report out the error before you get too deep into the code.  It will make future debugging easier if there is a problem as you'd catch it here instead of later when you try to use them.

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,144 @@
...
+  private final double ELEVATOR_KA = 0;
+  private final double reverseSoftLimit = -67;
+  private final double forwardSoftLimit = -1;
+
+  // Standard classes for controlling our elevator
+  private final ProfiledPIDController m_controller =
+      new ProfiledPIDController(
+          ELEVATOR_KP, ELEVATOR_KI, ELEVATOR_KD, new TrapezoidProfile.Constraints(2.45, 2.45));
+  ElevatorFeedforward m_feedforward =
+      new ElevatorFeedforward(ELEVATOR_KS, ELEVATOR_KG, ELEVATOR_KV, ELEVATOR_KA);
```
markpete, line 40: Can either m_controller or m_feedforward ever fail to initialize or return null?  If so, you might consider moving these to an initialization routine and then checking for null or the error there.  That way you can detect and alert on problems earlier and save yourself (or others) future debugging headaches.

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,144 @@
...
+    return routine.quasistatic(direction);
+  }
+
+  public Command sysIdDynamic(SysIdRoutine.Direction direction) {
+    return routine.dynamic(direction);
+  }
+
+  public void voltageDrive(Voltage drive) {
+    m_motor.setVoltage(drive.in(Units.Volts));
+    m_motor2.setVoltage(drive.in(Units.Volts));
```
markpete, line 101: Is it necessary to set this on m_motor2 given that it is in Follower mode from line 85?

braelynandthefrogs: Joseph said it's fine to keep it just in case

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
braelynandthefrogs: no.

braelynandthefrogs: words directly from James Dooley, so I'm deciding to trust them

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,147 @@
...
+      new ElevatorFeedforward(ELEVATOR_KS, ELEVATOR_KG, ELEVATOR_KV, ELEVATOR_KA);
+
+  private TalonFX m_motor;
+  private TalonFX m_motor2;
+
+  private final MutVoltage m_appliedVoltage = Units.Volts.mutable(0);
+  // Mutable holder for unit-safe linear distance values, persisted to avoid reallocation.
+  private final MutAngle m_angle = Units.Radians.mutable(0);
+  // Mutable holder for unit-safe linear velocity values, persisted to avoid reallocation.
+  private final MutAngularVelocity m_velocity = Units.RadiansPerSecond.mutable(0);
```
jamesdooley4, line 49: Should delete these since they're never used and likely won't be

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,147 @@
...
+  // This gearbox represents a gearbox containing 4 Vex 775pro motors.
+  private final double ELEVATOR_KP = 0.1;
+  private final double ELEVATOR_KI = 0;
+  private final double ELEVATOR_KD = 0;
+  private final double ELEVATOR_KS = 0;
+  private final double ELEVATOR_KV = 0;
+  private final double ELEVATOR_KG = 0;
+  private final double ELEVATOR_KA = 0;
+  private final double reverseSoftLimit = -67;
+  private final double forwardSoftLimit = -1;
```
jamesdooley4, line 33: Constants should be in ALL_CAPS, so consider REVERSE_SOFT_LIMIT and FORWARD_SOFT_LIMI

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,147 @@
...
+    var outputConfigs = new MotorOutputConfigs();
+    outputConfigs.NeutralMode = NeutralModeValue.Brake;
+    talonFXConfigurator.apply(outputConfigs);
+    talonFXConfigurator2.apply(outputConfigs);
+  }
+
+  public Command goUp() {
+    return startEnd(
+        () -> m_motor.set(-0.25), () -> m_motor.set(-0.02)); // test values from dev model bot
+  }
```
jamesdooley4, line 135: Would be best to define `-0.25` and `-0.02` above as constant variables & use those. Something like UP_VOLTAGE, DOWN_VOLTAGE and HOLD_VOLTAGE

jamesdooley4 (approved):

jamesdooley4 (approved):

PR #11, opened 2025-01-24
Started Climb Subsystem

iamawesomecat:
Made the subsystem file, made constants in hardware

### DIFF
--- a/src/main/java/frc/robot/Subsystems.java
+++ b/src/main/java/frc/robot/Subsystems.java
@@ -67,6 +67,7 @@ public class Subsystems {
       climbPivotSubsystem = new ClimbPivot();
     } else {
       climbPivotSubsystem = null;
+    }
 
     if (SPINNYCLAW_ENABLED) {
       spinnyClawSubsytem = new SpinnyClaw();
--- a/src/main/java/frc/robot/Subsystems.java
+++ b/src/main/java/frc/robot/Subsystems.java
@@ -4,13 +4,10 @@ import static frc.robot.Subsystems.SubsystemConstants.*;
 
 import frc.robot.generated.BonkTunerConstants;
 import frc.robot.subsystems.ArmPivot;
-
 import frc.robot.subsystems.ClimbPivot;
-
 import frc.robot.subsystems.DrivebaseWrapper;
 import frc.robot.subsystems.ElevatorSubsystem;
 import frc.robot.subsystems.VisionSubsystem;
-
 import frc.robot.subsystems.drivebase.CommandSwerveDrivetrain;
 
 public class Subsystems {
@@ -33,6 +30,7 @@ public class Subsystems {
   public final ElevatorSubsystem elevatorSubsystem;
   public final ArmPivot armPivotSubsystem;
   public final ClimbPivot climbPivotSubsystem;
+
   public Subsystems() {
     // Initialize subsystems here (don't forget to check if they're enabled!)
     // Add specification for bonk, Enum? get team number?
@@ -60,7 +58,7 @@ public class Subsystems {
       armPivotSubsystem = null;
     }
 
-    if (CLIMBPIVOT_ENABLED){
+    if (CLIMBPIVOT_ENABLED) {
       climbPivotSubsystem = new ClimbPivot();
     } else {
       climbPivotSubsystem = null;
--- a/src/main/java/frc/robot/Subsystems.java
+++ b/src/main/java/frc/robot/Subsystems.java
@@ -21,7 +21,7 @@ public class Subsystems {
 
     public static final boolean ELEVATOR_ENABLED = true;
     public static final boolean ARMPIVOT_ENABLED = true;
-    public static final boolean CLIMBPIVOT_ENABLED = true;
+    public static final boolean CLIMBPIVOT_ENABLED = false;
     public static final boolean SPINNYCLAW_ENABLED = true;
   }
 
--- a/src/main/java/frc/robot/Subsystems.java
+++ b/src/main/java/frc/robot/Subsystems.java
@@ -4,6 +4,7 @@ import static frc.robot.Subsystems.SubsystemConstants.*;
 
 import frc.robot.generated.BonkTunerConstants;
 import frc.robot.subsystems.ArmPivot;
+import frc.robot.subsystems.ClimbPivot;
 import frc.robot.subsystems.drivebase.CommandSwerveDrivetrain;
 
 public class Subsystems {
@@ -12,13 +13,14 @@ public class Subsystems {
 
     public static final boolean DRIVEBASE_ENABLED = true;
     public static final boolean ARMPIVOT_ENABLED = true;
+    public static final boolean CLIMBPIVOT_ENABLED = true;
   }
 
   // Subsystems go here
 
   public final CommandSwerveDrivetrain drivebaseSubsystem;
   public final ArmPivot armPivotSubsystem;
-
+  public final ClimbPivot climbPivotSubsystem;
   public Subsystems() {
     // Initialize subsystems here (don't forget to check if they're enabled!)
     // Add specification for bonk, Enum? get team number?
@@ -32,6 +34,12 @@ public class Subsystems {
       armPivotSubsystem = new ArmPivot();
     } else {
       armPivotSubsystem = null;
+    }
+
+    if (CLIMBPIVOT_ENABLED){
+      climbPivotSubsystem = new ClimbPivot();
+    } else {
+      climbPivotSubsystem = null;
     }
   }
 }
--- a/src/main/java/frc/robot/subsystems/ClimbPivot.java
+++ b/src/main/java/frc/robot/subsystems/ClimbPivot.java
@@ -33,7 +33,7 @@ public class ClimbPivot extends SubsystemBase {
     talonFXConfigurator2.apply(currentLimits);
   }
 
-  public Command MoveClimbMotor(double speed) {
+  public Command moveClimbMotor(double speed) {
     return runOnce(
         () -> {
           climbPivotMotorOne.set(speed);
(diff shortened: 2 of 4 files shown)


src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -0,0 +1,43 @@
...
+  private final TalonFX climbPivotMotorTwo;
+
+  public ClimbPivot() {
+    climbPivotMotorOne = new TalonFX(Hardware.CLIMB_PIVOT_MOTOR_ONE_ID);
+    climbPivotMotorTwo = new TalonFX(Hardware.CLIMB_PIVOT_MOTOR_TWO_ID);
+
+    configureMotor();
+  }
+
+  public void configureMotor() {
```
markpete, line 21: Should this be private instead of public?  Do you need this externally callable?

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -0,0 +1,43 @@
...
+    // enable stator current limit
+    currentLimits.StatorCurrentLimit = 5;
+    currentLimits.StatorCurrentLimitEnable = true;
+    currentLimits.SupplyCurrentLimit = 5;
+    currentLimits.SupplyCurrentLimitEnable = true;
+    talonFXConfigurator.apply(currentLimits);
+    talonFXConfigurator2.apply(currentLimits);
+  }
+
+  public Command MoveClimbMotor(double speed) {
```
KangarooKoala, line 36:
```suggestion
  public Command moveClimbMotor(double speed) {
```

KangarooKoala (commented):
We should also add controls bindings (which also include stopping the motors), though that doesn't need to happen if we want to merge this PR before Aarya arrives during the meeting at 6:30 to pick what the bindings should be.

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -0,0 +1,43 @@
...
+  private final TalonFX climbPivotMotorTwo;
+
+  public ClimbPivot() {
+    climbPivotMotorOne = new TalonFX(Hardware.CLIMB_PIVOT_MOTOR_ONE_ID);
+    climbPivotMotorTwo = new TalonFX(Hardware.CLIMB_PIVOT_MOTOR_TWO_ID);
+
+    configureMotor();
+  }
+
+  private void configureMotor() {
```
KangarooKoala, line 21:
```suggestion
  private void configureMotors() {
```
Oops, missed this earlier.

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -0,0 +1,43 @@
...
+public class ClimbPivot extends SubsystemBase {
+
+  private final TalonFX climbPivotMotorOne;
+  private final TalonFX climbPivotMotorTwo;
+
+  public ClimbPivot() {
+    climbPivotMotorOne = new TalonFX(Hardware.CLIMB_PIVOT_MOTOR_ONE_ID);
+    climbPivotMotorTwo = new TalonFX(Hardware.CLIMB_PIVOT_MOTOR_TWO_ID);
+
+    configureMotor();
```
KangarooKoala, line 18:
```suggestion
    configureMotors();
```

KangarooKoala (commented):
For posterity- Since Aarya can't come to the meeting today, we can do the controls later.

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -0,0 +1,43 @@
...
+
+    var currentLimits = new CurrentLimitsConfigs();
+
+    // enable stator current limit
+    currentLimits.StatorCurrentLimit = 5;
+    currentLimits.StatorCurrentLimitEnable = true;
+    currentLimits.SupplyCurrentLimit = 5;
+    currentLimits.SupplyCurrentLimitEnable = true;
+    talonFXConfigurator.apply(currentLimits);
+    talonFXConfigurator2.apply(currentLimits);
```
jamesdooley4, line 33: Need to also set the motors into brake mode

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  public Command moveClimbMotor(double speed) {
+    return runOnce(
```
jamesdooley4, line 37: Would be better to use `run( ... ).finallyDo( ... )` so the motor stops when this command ends.

jamesdooley4 (changes requested):

src/main/java/frc/robot/Subsystems.java
```diff
@@ -47,16 +50,24 @@ public Subsystems() {
...
       armPivotSubsystem = new ArmPivot();
     } else {
       armPivotSubsystem = null;
     }
+
+    if (CLIMBPIVOT_ENABLED) {
+      climbPivotSubsystem = new ClimbPivot();
+    } else {
+      climbPivotSubsystem = null;
+
```
KangarooKoala, line 70:
```suggestion
      climbPivotSubsystem = null;
    }

```

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -0,0 +1,53 @@
...
+    var talonFXMotorOutput = new MotorOutputConfigs();
+
+    var currentLimits = new CurrentLimitsConfigs();
+
+    // enable stator current limit
+    currentLimits.StatorCurrentLimit = 5;
+    currentLimits.StatorCurrentLimitEnable = true;
+    currentLimits.SupplyCurrentLimit = 5;
+    currentLimits.SupplyCurrentLimitEnable = true;
+    talonFXMotorOutput.withNeutralMode(NeutralModeValue.Brake);
```
KangarooKoala, line 35:
```suggestion
    // Enable brake mode
    talonFXMotorOutput.NeutralMode = NeutralModeValue.Brake;
```
The convention is that you use the return value of methods that start with `with` and side effects are unimportant. (e.g., `currentLimits = new CurrentLimitsConfigs().withNeutralMode(NeutralModeValue.Brake)`)

src/main/java/frc/robot/Subsystems.java
```diff
@@ -20,6 +21,7 @@ public static class SubsystemConstants {
 
     public static final boolean ELEVATOR_ENABLED = true;
     public static final boolean ARMPIVOT_ENABLED = true;
+    public static final boolean CLIMBPIVOT_ENABLED = true;
```
jamesdooley4, line 24: Until the climb pivot is actually installed on the robot, please set this to `false`

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -0,0 +1,54 @@
...
+
+    configureMotors();
+  }
+
+  private void configureMotors() {
+    var talonFXConfigurator = climbPivotMotorOne.getConfigurator();
+    var talonFXConfigurator2 = climbPivotMotorTwo.getConfigurator();
+    var talonFXMotorOutput = new MotorOutputConfigs();
+
+    var currentLimits = new CurrentLimitsConfigs();
```
jamesdooley4, line 28: Would be better to move these down to immediately before each of these variables is first used so if code that uses them is deleted then this call to allocate & construct them is more likely to also be deleted.

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -0,0 +1,54 @@
...
+
+    // enable stator current limit
```
jamesdooley4, line 30: This does more than just the stator current limit, so suggest 'set and enable current limits'

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -0,0 +1,54 @@
...
+  private final TalonFX climbPivotMotorTwo;
+
+  public ClimbPivot() {
+    climbPivotMotorOne = new TalonFX(Hardware.CLIMB_PIVOT_MOTOR_ONE_ID);
+    climbPivotMotorTwo = new TalonFX(Hardware.CLIMB_PIVOT_MOTOR_TWO_ID);
+
+    configureMotors();
+  }
+
+  private void configureMotors() {
```
jamesdooley4, line 23: Also need to set one of the motors to follow the other, as I expect they'll be e mechanically linked

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -12,42 +14,50 @@ public class ClimbPivot extends SubsystemBase {
 
   private final TalonFX climbPivotMotorOne;
   private final TalonFX climbPivotMotorTwo;
+  private final DigitalInput climbSensor;
+
+  public static boolean IS_CLIMBED = false;
```
KangarooKoala, line 19: Where is this used? Should it be `final`?

iamawesomecat: Not final, will change to true when the sensor detects the cage (thats what the sensor is supposed to do, right?)

Can be used in a sequential, has no real use yet.

KangarooKoala: Why not just call `checkClimbSensor()` instead of checking that variable? And if it's meant to be changed, why is it named like a constant?

iamawesomecat: Oh, yeah that makes sense. Removed the variable

KangarooKoala (commented):
Not sure these are the controls we want for manual control (I was thinking of one button moves it out, other button moves it in), but otherwise looks good.

jamesdooley4 (approved):

PR #16, opened 2025-01-25
ArmSensor, and other Sensor work.

Skymouse888:


jamesdooley4 (approved):

src/main/java/frc/robot/sensors/ArmSensor.java
```diff
@@ -0,0 +1,37 @@
...
+import static edu.wpi.first.units.Units.Millimeter;
+
+import au.grapplerobotics.ConfigurationFailedException;
+import au.grapplerobotics.LaserCan;
+import edu.wpi.first.units.measure.Distance;
+import frc.robot.Hardware;
+
+public class ArmSensor {
+
+  private final LaserCan mainSensor;
```
markpete, line 12: Recommend a more descriptive name for ease of debugging, e.g. mainLaserCanSensor or similar

src/main/java/frc/robot/sensors/ArmSensor.java
```diff
@@ -0,0 +1,37 @@
...
+      System.out.println("Configuration Failed! " + e);
+    }
+  }
+
+  public Distance getSensorDistance() {
+    LaserCan.Measurement measurement = mainSensor.getMeasurement();
+    if (measurement != null && measurement.status == LaserCan.LASERCAN_STATUS_VALID_MEASUREMENT) {
+      return Millimeter.of(measurement.distance_mm);
+    } else {
+      return Millimeter.of(-1);
```
markpete, line 34: Should we just return -1 here instead of Millimeter.of(-1)?  Does that mean you have to decode it later?

markpete, line 34: Is it useful to return a different result for measurement == null (-2) vs. LaserCan.LASERCAN_STATUS_VALID*  (-1)?

E.g. if the configuration failed vs. an invalid measurement?

PR #23, opened 2025-02-07
Elevator Reset Zero Position Command

braelynandthefrogs:
created and works with an error of about .04

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -100,12 +100,24 @@ public class Controls {
     // Controls binding goes here
     operatorController.leftTrigger().whileTrue(s.elevatorSubsystem.goUpPower());
     operatorController.rightTrigger().whileTrue(s.elevatorSubsystem.goDownPower());
-    operatorController.y().onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_FOUR_POS));
-    operatorController.x().onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_THREE_POS));
-    operatorController.b().onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_TWO_POS));
-    operatorController.a().onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_ONE_POS));
+    operatorController
+        .y()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_FOUR_POS).withName("L4"));
+    operatorController
+        .x()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_THREE_POS).withName("L3"));
+    operatorController
+        .b()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_TWO_POS).withName("L2"));
+    operatorController
+        .a()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_ONE_POS).withName("L1"));
+    operatorController
+        .rightBumper()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.INTAKE).withName("IntakePos"));
     operatorController.povUp().whileTrue(s.elevatorSubsystem.goUp());
     operatorController.povDown().whileTrue(s.elevatorSubsystem.goDown());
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -36,8 +36,8 @@ public class ElevatorSubsystem extends SubsystemBase {
   private final double ELEVATOR_KS = 0;
   private final double ELEVATOR_KV = 0;
   private final double ELEVATOR_KA = 0;
-  private final double REVERSE_SOFT_LIMIT = STOWED - 1; // soft limits arent currently working
-  private final double FORWARD_SOFT_LIMIT = LEVEL_FOUR_POS + 3;
+  private final double REVERSE_SOFT_LIMIT = INTAKE - 0.05; // soft limits arent currently working
```
KangarooKoala, line 39: Are the soft limits still not working? (I'm guessing we just forgot to update this comment)

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -24,9 +24,9 @@ public class ElevatorSubsystem extends SubsystemBase {
...
   public static final double LEVEL_THREE_POS = 6;
   public static final double LEVEL_TWO_POS = 4;
-  public static final double LEVEL_ONE_POS = 1;
+  public static final double LEVEL_ONE_POS = 2;
   public static final double STOWED = 1;
-  public static final double INTAKE = 1;
+  public static final double INTAKE = 0;
   public static final double MANUAL = 1;
   private static final double POS_TOLERANCE = 0.02;
   // This gearbox represents a gearbox containing 4 Vex 775pro motors.
```
KangarooKoala, line 32: We should remove this stale comment at some point.

KangarooKoala (commented):
Just some random thoughts when looking through code in the morning. (Not related to the PR changes)

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
braelynandthefrogs: yes, they are working now, just forgot to update the comment. thanks for noticing, i will fix that

src/main/java/frc/robot/Controls.java
```diff
@@ -100,12 +100,24 @@ private void configureElevatorBindings() {
...
+        .x()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_THREE_POS).withName("L3"));
+    operatorController
+        .b()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_TWO_POS).withName("L2"));
+    operatorController
+        .a()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_ONE_POS).withName("L1"));
+    operatorController
+        .rightBumper()
```
KangarooKoala, line 116: Note for future reference- This does conflict with the spinny claw controls. This is fine for the test benches since they won't both be enabled, but this needs to be fixed before we run on the bot (when that gets done).

src/main/java/frc/robot/Controls.java
```diff
@@ -100,12 +100,24 @@ private void configureElevatorBindings() {
...
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_FOUR_POS).withName("L4"));
+    operatorController
+        .x()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_THREE_POS).withName("L3"));
+    operatorController
+        .b()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_TWO_POS).withName("L2"));
+    operatorController
+        .a()
+        .onTrue(s.elevatorSubsystem.setLevel(ElevatorSubsystem.LEVEL_ONE_POS).withName("L1"));
```
jamesdooley4, line 114: For the command names, please prefix with `Elevator`, such as `ElevatorL4` and `ElevatorIntakePos`. That'll disambiguate this from the 'L4' preset command I expect will also exist for the arm/elbow.

jamesdooley4 (dismissed):

src/main/java/frc/robot/Controls.java
braelynandthefrogs: yep, I was aware and knew it wouldn't be a problem currently since we won't be running both at the same time currently

PR #25, opened 2025-02-11
Climb log

iamawesomecat:
info in commit descriptions

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -141,15 +141,8 @@ public class Controls {
     if (s.climbPivotSubsystem == null) {
       return;
     }
-
-    operatorController
-        .start()
-        .onTrue(
-            s.climbPivotSubsystem
-                .moveClimbMotor(0.1)
-                .withTimeout(0.5)
-                .andThen(s.climbPivotSubsystem.moveClimbMotor(-0.1))
-                .withTimeout(0.5));
+    // Idk if this is great code or horrible code
+    operatorController.start().onTrue(s.climbPivotSubsystem.toggleClimb());
   }
 
   private void configureSpinnyClawBindings() {
--- a/src/main/java/frc/robot/subsystems/ClimbPivot.java
+++ b/src/main/java/frc/robot/subsystems/ClimbPivot.java
@@ -102,7 +102,7 @@ public class ClimbPivot extends SubsystemBase {
     return climbSensor.get();
   }
 
-  public double getClimbVelocity(){
+  public double getClimbVelocity() {
     return climbPivotMotorOne.getVelocity().getValueAsDouble();
   }
 
@@ -127,7 +127,9 @@ public class ClimbPivot extends SubsystemBase {
     shuffleboardTab
         .addBoolean("Cage Detected", () -> checkClimbSensor())
         .withWidget(BuiltInWidgets.kBooleanBox);
-    shuffleboardTab.addDouble("Motor Speed", () -> getClimbVelocity()).withWidget(BuiltInWidgets.kTextView);
+    shuffleboardTab
+        .addDouble("Motor Speed", () -> getClimbVelocity())
+        .withWidget(BuiltInWidgets.kTextView);
   }
 
   @Override
--- a/src/main/java/frc/robot/subsystems/ClimbPivot.java
+++ b/src/main/java/frc/robot/subsystems/ClimbPivot.java
@@ -7,6 +7,7 @@ import com.ctre.phoenix6.hardware.TalonFX;
 import com.ctre.phoenix6.signals.NeutralModeValue;
 import edu.wpi.first.wpilibj.DigitalInput;
 import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.Commands;
 import edu.wpi.first.wpilibj2.command.SubsystemBase;
 import frc.robot.Hardware;
 
@@ -15,6 +16,14 @@ public class ClimbPivot extends SubsystemBase {
   private final TalonFX climbPivotMotorOne;
   private final TalonFX climbPivotMotorTwo;
   private final DigitalInput climbSensor;
+
+  private final double CLIMB_OUT_PRESET = 90;
+  private final double CLIMB_IN_PRESET = 0;
+  private final double CLIMB_IN_SPEED = 0.5;
+  private final double CLIMB_OUT_SPEED = -0.5;
+  private boolean isClimbOut = false;
+  private boolean isClimbIn = true;
+  private boolean nextMoveOut = true;
 
   public ClimbPivot() {
     climbPivotMotorOne = new TalonFX(Hardware.CLIMB_PIVOT_MOTOR_ONE_ID);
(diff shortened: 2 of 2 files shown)


src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -55,7 +72,77 @@ public Command moveClimbMotor(double speed) {
...
+
   public boolean checkClimbSensor() {
     return climbSensor.get();
   }
+
+  public double getClimbVelocity() {
+    return climbPivotMotorOne.getVelocity().getValueAsDouble();
+  }
+
+  public void logging() {
```
jamesdooley4, line 109: Would be better named something like `setupLogging`

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -55,7 +72,77 @@ public Command moveClimbMotor(double speed) {
...
+        .addBoolean("Cage Detected", () -> checkClimbSensor())
+        .withWidget(BuiltInWidgets.kBooleanBox);
+    shuffleboardTab
+        .addDouble("Motor Speed", () -> getClimbVelocity())
+        .withWidget(BuiltInWidgets.kTextView);
+  }
+
+  @Override
+  public void periodic() {
+    if (climbPivotMotorOne.getPosition().getValueAsDouble() == CLIMB_OUT_PRESET) {
```
jamesdooley4, line 137: This will likely never be true, as it requires the motor to be within about `2.220446049250313E-16` rotations of the preset value. It'd be better to compare the position with the target value and set `isClimbOut` if it's "close" by some definition. Other subsystems define a 'tolerance' value to specify what "close" means.

jamesdooley4 (changes requested):

src/main/java/frc/robot/subsystems/ClimbPivot.java
iamawesomecat: Sounds good, I just named it logging because it was consistently getting updated through the lambda's in the shuffleboard entries, but Ill change it to that 👍

src/main/java/frc/robot/subsystems/ClimbPivot.java
iamawesomecat: Yeah I was thinking about that. Ill implement that now

jamesdooley4 (approved):

src/main/java/frc/robot/Controls.java
```diff
@@ -141,15 +141,8 @@ private void configureClimbPivotBindings() {
...
-    operatorController
-        .start()
-        .onTrue(
-            s.climbPivotSubsystem
-                .moveClimbMotor(0.1)
-                .withTimeout(0.5)
-                .andThen(s.climbPivotSubsystem.moveClimbMotor(-0.1))
-                .withTimeout(0.5));
+    // Idk if this is great code or horrible code
+    operatorController.start().onTrue(s.climbPivotSubsystem.toggleClimb());
```
markpete, line 145: When does toggleClimb() return true and when does it return false?  You want to make sure this has a well-defined ending condition and doesn't get stuck here in an infinite loop toggling climb on and off.

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -55,7 +74,79 @@ public Command moveClimbMotor(double speed) {
             });
   }
 
+  public Command toggleClimb() {
+    return Commands.either(
```
markpete, line 78: If both the climbIn and climbOut are running in parallel here and it's waiting for either one to complete, there is going to be a race condition.  nextMoveOut will be true or false depending on which routine randomly gets there last and the motor will be fighting against itself to CLIMB_IN vs. CLIMB_OUT.

If the intent here is to toggle you need to keep the state in a global variable and then pick the action based on the state.

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -55,7 +74,79 @@ public Command moveClimbMotor(double speed) {
...
+            startEnd(
+                    () -> {
+                      // climb out
+                      nextMoveOut = false;
+                      climbPivotMotorOne.set(CLIMB_OUT_SPEED);
+                    },
+                    () -> {
+                      climbPivotMotorOne.stopMotor();
```
markpete, line 86: Should we add periodic() here?

src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -55,7 +74,79 @@ public Command moveClimbMotor(double speed) {
...
+                    })
+                .until(() -> isClimbOut),
+            startEnd(
+                    () -> {
+                      // climb in
+                      nextMoveOut = true;
+                      climbPivotMotorOne.set(CLIMB_IN_SPEED);
+                    },
+                    () -> {
+                      climbPivotMotorOne.stopMotor();
```
markpete, line 96: Should we add periodic() here?  Idea being make sure it gets called and checked in the loop?

src/main/java/frc/robot/Controls.java
KangarooKoala: The way command durations work is as follows:
* Internally, each command implements a `isFinished()` method which the [CommandScheduler](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/CommandScheduler.html)\* checks every code loop to determine when to remove the command from the set of running commands.
* WPILib provides factory methods which return commands with particular end conditions. For example, `runOnce(() -> action())` returns a command that runs the given action when the command starts and which finishes instantly (`isFinished()` always returns true), `run(() -> action())` returns a command that runs the given action every code loop while the command is running and which never finishes (`isFinished()` always returns false), and `waitUntil(() -> condition())` returns a command that does nothing and which finishes when `condition()` returns true. (`isFinished()` returns `condition()`)

For more information about the available command factories, look for the methods returning `Command` in [`Subsystem`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Subsystem.html) or [`Commands`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Commands.html).

src/main/java/frc/robot/subsystems/ClimbPivot.java
KangarooKoala: > If both the climbIn and climbOut are running in parallel here and it's waiting for either one to complete, there is going to be a race condition.

[`Commands.either()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Commands.html#either(edu.wpi.first.wpilibj2.command.Command,edu.wpi.first.wpilibj2.command.Command,java.util.function.BooleanSupplier)) only runs one of the commands at a time. (The command factories with the describe semantics are [`Commands.parallel()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Commands.html#parallel(edu.wpi.first.wpilibj2.command.Command...)) and [`command.alongWith()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Command.html#alongWith(edu.wpi.first.wpilibj2.command.Command...)). There are also [`Commands.deadline()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Commands.html#deadline(edu.wpi.first.wpilibj2.command.Command,edu.wpi.first.wpilibj2.command.Command...)), [`command.deadlineFor()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Command.html#deadlineFor(edu.wpi.first.wpilibj2.command.Command...)), [`Commands.race()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Commands.html#race(edu.wpi.first.wpilibj2.command.Command...)), and [`command.raceWith()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Command.html#raceWith(edu.wpi.first.wpilibj2.command.Command...)) factories which run commands in parallel but have different end conditions.

> If the intent here is to toggle you need to keep the state in a global variable and then pick the action based on the state.

Yep! That's what the current implementation does. (`nextMoveOut` is the global state and `Commands.either(..., ..., () -> nextMoveOut)` is what picks the action based on the state)

src/main/java/frc/robot/subsystems/ClimbPivot.java
KangarooKoala: No- `periodic()` is called every single code loop, and we do not want to stop the motor continuously. [`subsystem.startEnd()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Subsystem.html#startEnd(java.lang.Runnable,java.lang.Runnable)) runs the first action when the command starts and runs the section action (in this case, `climbPivotMotorOne.stopMotor()`) when the command finishes (and the timing is determined by the [`command.until()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Command.html#until(java.util.function.BooleanSupplier)) [decorator method](https://en.wikipedia.org/wiki/Decorator_pattern)).

PR #26, opened 2025-02-12
Elevator Limit Switch - untested

braelynandthefrogs:


src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -44,8 +46,11 @@ public class ElevatorSubsystem extends SubsystemBase {
   private final MotionMagicVoltage m_request = new MotionMagicVoltage(0);
 
   // motors
-  private TalonFX m_motor;
-  private TalonFX m_motor2;
+  private final TalonFX m_motor;
+  private final TalonFX m_motor2;
```
markpete, line 50: Can we name these something more clear?  LIke m_motor_left and m_motor_right?  Some way that people will know which is which on the bot.

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -44,8 +46,11 @@ public class ElevatorSubsystem extends SubsystemBase {
...
+
+  // limit switch
+  private final DigitalInput input = new DigitalInput(Hardware.ELEVATOR_LIMIT_SWITCH);
```
markpete, line 53: need a better variable name here as well, e.g. m_elevator_limit_input

iamawesomecat:
This branch should probably be closed @Jetblackdragon @KangarooKoala thoughts?

PR #34, opened 2025-02-20
2/18-2/19 arm and superstructure

### LINKED ISSUE #28: Drivebase acceleration limit

KangarooKoala:
Notable happenings, and lessons (for posterity):
* SysId did not work for the arm pivot (The arm pivot is likely too light and has too short of a travel), so we ended up using https://reca.lc and some guessing adjustments to kG and kS- Don't whack everything with the same hammer
* We spent a lot of time figuring out arm pivot and spinny claw issues before realizing they needed to be inverted- Verify directions as soon as the physical setup is received
* Debugging was slowed down by having to add names to commands- Require from the beginning of build season that all commands have names
* We needed to configure the arm encoder offset (in Phoenix Tuner) so that 0 is horizontal, since that's what `GravityTypeValue.Arm_Cosine` assumes- Verify assumptions of feedforwards
* The first test of the L4 superstructure preset (done on 56208d0cf3135a24d2b478e89b72cc429d16f11f, maybe with some changes) was done with a coral on the reef, and the arm didn't let go of the coral as it went down and ended up breaking the reef branch- Verify sequence logic away from field elements and think about potential collisions
* In a subsequent test of the L4 superstructure preset, the arm pivot (still holding the coral) swung down while the elevator was retracting, leading to the coral jamming between the elevator crossbraces- Verify sequence logic without game pieces and think about potential collisions; Pick designs that avoid complicated "Keep out" portions of the state space (including the presence of game pieces)
* During testing, after reenabling the elevator went to L4 with a coral in the trough, leading to the coral hitting the bottom elevator crossbrace- Pick designs that avoid complicated "Keep out" portions of the state space (including the presence of game pieces); Ensure motor commanded positions are cleared during disable to avoid unexpected behavior
* During drive practice, we went to L4 with a coral in the trough that hadn't been transferred to the arm pivot, leading to the coral ramming into the bottom elevator crossbrace- Pick designs that avoid complicated "Keep out" portions of the state space (including the presence of game pieces); Use sensors (when attached) to prevent bad transitions between states

Fixes #28.

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -8,6 +8,7 @@ import edu.wpi.first.math.kinematics.ChassisSpeeds;
 import edu.wpi.first.wpilibj.GenericHID.RumbleType;
 import edu.wpi.first.wpilibj2.command.Commands;
 import edu.wpi.first.wpilibj2.command.button.CommandXboxController;
+import edu.wpi.first.wpilibj2.command.button.RobotModeTriggers;
 import edu.wpi.first.wpilibj2.command.button.Trigger;
 import edu.wpi.first.wpilibj2.command.sysid.SysIdRoutine.Direction;
 import frc.robot.generated.BonkTunerConstants;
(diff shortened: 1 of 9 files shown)


KangarooKoala:
Merging now to keep the days clean- Made #35 to track the comments

PR #32, opened 2025-02-17
replaced SuperStructureBindings with BranchHeight shenanigans

SnappyRiffs:

PR #36, opened 2025-02-20
2/20 preset testing

KangarooKoala:
Implement operator queueing/mode selection controls (Operator selects next level, driver tells robot when to move to that level)
Makes the operator manual elevator up and down controls continuous instead of discrete on/off
Fixes motor position terminology in arm pivot
Fixes L2 and L3 presets
Mostly fixes acceleration limits

Closes #32 (because this contains the same contents and a fix for a bug).

About the acceleration limits- The old code was incorrect (it would apply the pre-limit acceleration and it wouldn't handle negative rotation speeds correctly). However, after fixing it, we discovered that applying the deadband limits after the acceleration limits causes trouble (since the acceleration should map certain values to within the deadband that should not be set to 0). The fix is to apply the deadbands before the acceleration limit, but we can't do that without implementing deadband ourselves, which was proving difficult. So far we haven't had issues with tipping (even with elevator up), so both features are on low priority for the time being.

Side note about the operator queueing/mode selection controls- This presents a use case where pressing a button again should cause the command to be rescheduled. However, when a command is scheduled, the scheduler will skip it if the command is currently running. To get around this, we use the (unspecified) fact that bindings are always checked in the order that they are made so that when the trigger is pressed, first we schedule a different command with overlapping requirements to cancel the superstructure command and then schedule the actual superstructure command to start it again.
In hindsight, we could avoid relying on the unspecified definition of the binding execution order by using `interruptor.asProxy().andThen(actualCommand.asProxy())` (assuming `interruptor` finishes instantly) or `new ScheduleCommand(interruptor).andThen(new ScheduleCommand(actualCommand))` (straying from strict adherence to command factories and using a rarely used command). However, that hasn't been tested (and notably, it would add one command scheduler poll of delay).

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -4,6 +4,7 @@ import static edu.wpi.first.units.Units.*;
 
 import com.ctre.phoenix6.swerve.SwerveModule.DriveRequestType;
 import com.ctre.phoenix6.swerve.SwerveRequest;
+import edu.wpi.first.math.MathUtil;
 import edu.wpi.first.math.kinematics.ChassisSpeeds;
 import edu.wpi.first.wpilibj.GenericHID.RumbleType;
 import edu.wpi.first.wpilibj2.command.Commands;
@@ -159,6 +160,7 @@ public class Controls {
     operatorController
         .a()
         .onTrue(superStructure.levelOne(driverController.rightBumper()).withName("level 1"));
+    operatorController.rightBumper().onTrue(superStructure.stow().withName("Stow"));
     driverController.a().onTrue(superStructure.intake());
     if (sensors.armSensor != null) {
       sensors.armSensor.inTrough().onTrue(superStructure.intake());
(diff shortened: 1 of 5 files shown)


src/main/java/frc/robot/Controls.java
```diff
@@ -102,15 +107,22 @@ private void configureDrivebaseBindings() {
...
+                  //     translationLimit.times(
+                  //         Math.min(
+                  //             1,
+                  //             Math.abs(
+                  //                 MAX_ROTATION_ACCELERATION
+                  //                     / translationLimit.omegaRadiansPerSecond)));
+                  // This *should* be rotationLimit, but the acceleration limiting causes the
+                  // commanded speed to fall into the deadband. The proper fix is to do the deadband
+                  // first, which relies on us doing the deadband ourselves, which is being
+                  // difficult.
```
jamesdooley4, line 125: Generally, I've seen the deadband applied on the input value (the `driverController` inputs above) with 
[MathUtil.applyDeadband](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/MathUtil.html#applyDeadband(double,double)) to avoid issues like this.

jamesdooley4 (approved):

PR #38, opened 2025-02-22
2/21 testing

iamawesomecat:
tested a basic l1 score
- Wasnt entirely accurate, either due to lack of bumpers or due to the strange orientation. More research could be done, but not a big deal. Low priority? 

Tested climb 🔥🔥🔥🔥🔥
- It worked well, mech recommended that we have a separate climb in position that is different from the starting configuration, this way the chain doesnt touch the elevator. Since we are roughly 8 pounds lighter than the weight limit (accounting for the things mech is yet to add), we have some weight we can use as ballast.

Investigated removing Algae
- Determined that we need separate presets to do it on l2 and l3.

### DIFF
--- a/src/main/java/frc/robot/subsystems/SuperStructure.java
+++ b/src/main/java/frc/robot/subsystems/SuperStructure.java
@@ -60,23 +60,16 @@ public class SuperStructure {
         stow());
   }
 
-  public Command levelOnePrescore() {
-    return Commands.parallel(
-        elevator.setLevel(ElevatorSubsystem.LEVEL_ONE_POS),
-        armPivot.moveToPosition(ArmPivot.PRESET_L1),
-        spinnyClaw.stop());
-  }
-
-  public Command levelOneScore() {
-    return Commands.parallel(
-        elevator.setLevel(ElevatorSubsystem.STOWED),
-        armPivot.moveToPosition(ArmPivot.PRESET_L1),
-        spinnyClaw.extakePower());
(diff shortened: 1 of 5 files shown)


src/main/java/frc/robot/subsystems/SuperStructure.java
```diff
@@ -60,23 +60,16 @@ public Command levelTwo(BooleanSupplier score) {
...
     return Commands.sequence(
-        levelOnePrescore(), Commands.waitUntil(score), levelOneScore(), stow());
+        Commands.parallel(
+            elevator.setLevel(ElevatorSubsystem.LEVEL_ONE_POS),
+            armPivot.moveToPosition(ArmPivot.PRESET_L1),
+            spinnyClaw.stop()),
+        Commands.waitUntil(score),
+        elevator.setLevel(ElevatorSubsystem.LEVEL_TWO_POS),
+        spinnyClaw.holdExtakePower().withTimeout(0.15),
+        stow());
```
KangarooKoala, line 72: Later we should have separate constants for `LEVEL_ONE_PRE_POS` and `LEVEL_ONE_POS` and use them.

KangarooKoala (approved):
Looks good to me, though I'd like to wait to see James' review. Note to self- At some point before competition we should clean up outdated comments (some from today's changes, also some from earlier).

jamesdooley4 (approved):

PR #41, opened 2025-02-25
Slow mode

iamawesomecat:
Will likely need to be reimplemented for deadband maybe? unsure


src/main/java/frc/robot/Controls.java
```diff
@@ -101,11 +102,16 @@ private void configureDrivebaseBindings() {
                   ChassisSpeeds targetSpeeds =
                       new ChassisSpeeds(
                           -driverController.getLeftY()
-                              * MaxSpeed, // Drive forward with negative Y (forward)
+                              * MaxSpeed
+                              * (slowMode ? 0.5 : 1), // Drive forward with negative Y (forward)
```
KangarooKoala, line 106: Check `driverController.start().getAsBoolean()` instead of `slowMode`.

Side note (mostly for James)- Controller button triggers are cached, so this doesn't cause unnecessary allocations. We could also store it as a `BooleanSupplier` in the class so that it's guaranteed not reallocated.

src/main/java/frc/robot/Controls.java
```diff
@@ -101,11 +101,20 @@ private void configureDrivebaseBindings() {
                   ChassisSpeeds targetSpeeds =
                       new ChassisSpeeds(
                           -driverController.getLeftY()
-                              * MaxSpeed, // Drive forward with negative Y (forward)
+                              * MaxSpeed
+                              * (driverController.start().getAsBoolean()
+                                  ? 0.5
+                                  : 1), // Drive forward with negative Y (forward)
```
jamesdooley4, line 107: Would be better to read `driverController.start()` once inside the lambda, store the result (0.5 or 1.0) in a local variable and then use that variable in each of these as a multiplier for the max speed or max angular rate

src/main/java/frc/robot/Controls.java
```diff
@@ -101,11 +101,20 @@ private void configureDrivebaseBindings() {
...
                           -driverController.getLeftX()
-                              * MaxSpeed, // Drive left with negative X (left)
+                              * MaxSpeed
+                              * (driverController.start().getAsBoolean()
+                                  ? 0.5
+                                  : 1), // Drive left with negative X (left)
```
jamesdooley4, line 112: Along with, or instead of simply modifying the maximum value, consider squaring or cubing the driverController output so the same stick movement results in smaller resulting values. If squaring, be sure to separately add in the sign of the original output value. If cubing, no need for that.

jamesdooley4: This is a trick I've seen used to let the drive controller still have full range of speed / rotation, while letting small stick motions be more precise and large stick movements be more granular.

jamesdooley4 (approved):

KangarooKoala:
This PR might need to include shifting to a (per-axis) WPILib deadband instead of the CTRE one so that we can do the deadband before the scaling and optionally squaring or cubing steps. (Otherwise, the deadband grows- Scaling the input by 0.5 expands the deadband from 0.1 to 0.2, squaring expands it from 0.1 to √(0.1) ≈ 0.32, and cubing expands it from 0.1 to ∛(0.1) ≈ 0.46.)

src/main/java/frc/robot/Controls.java
jamesdooley4: FYI, the comment here is more about seeing the exact same block of code 3 times, and less about any computation / memory overhead from making the exact same set of calls 3 times.

src/main/java/frc/robot/Controls.java
```diff
@@ -90,6 +88,10 @@ private void configureDrivebaseBindings() {
       // Stop running this method
       return;
     }
+
+    double getLeftX = Math.pow(MathUtil.applyDeadband(driverController.getLeftX(), 0.1), 3);
+    double getLeftY = Math.pow(MathUtil.applyDeadband(driverController.getLeftY(), 0.1), 3);
+    double getRightX = Math.pow(MathUtil.applyDeadband(driverController.getRightX(), 0.1), 3);
```
jamesdooley4: Make sure to test these with Strummer before checking in. Searching https://github.com/search?q=driverController.getLeftX+AND+Math.pow&type=code I see some teams doing a cube here, some doing a square, and from https://github.com/search?q=driverController.getLeftX&type=code some just use the value as-is.
And [the Technotoes only cube the rotation value](https://github.com/technototes/IntoTheDeep2024/blob/6ec6b0732f9900c53915d7b13738b49e098a1901/Hoops/src/main/java/org/firstinspires/ftc/hoops/commands/JoystickDriveCommand.java#L52), and not the translation inputs.

jamesdooley4:
@iamawesomecat - please resolve merge conflicts with main (ideally by merging main into this branch)

src/main/java/frc/robot/Controls.java
```diff
@@ -134,11 +144,19 @@ private void configureDrivebaseBindings() {
...
-        .applyRequest(() -> brake)
-        .ignoringDisable(true)
-        .withName("Brake")
-        .schedule();
+    RobotModeTriggers.disabled()
+        .onTrue(
+            s.drivebaseSubsystem
+                .applyRequest(() -> brake)
+                .ignoringDisable(true)
+                .withName("Disable Brake"));
```
jamesdooley4, line 152: ? This sets the robot in brake mode, so either the comment is wrong or this is calling the wrong command.

jamesdooley4: If the comment is wrong, I think this also won't do anything until the robot becomes enabled since motors can't be actuated during disabled mode, so the wheels won't move to the X position until the robot is enabled (and assuming nothing else overrides the swerve command set here before the robot is enabled).

src/main/java/frc/robot/Subsystems.java
```diff
@@ -20,9 +21,9 @@ public static class SubsystemConstants {
 
     public static final boolean DRIVEBASE_ENABLED = true;
     public static final boolean VISION_ENABLED = false;
-    public static final boolean ELEVATOR_ENABLED = true;
-    public static final boolean ARMPIVOT_ENABLED = true;
-    public static final boolean SPINNYCLAW_ENABLED = true;
+    public static final boolean ELEVATOR_ENABLED = false;
+    public static final boolean ARMPIVOT_ENABLED = false;
+    public static final boolean SPINNYCLAW_ENABLED = false;
```
KangarooKoala, line 26: These subsystems should stay enabled.

src/main/java/frc/robot/Controls.java
```diff
@@ -134,11 +144,19 @@ private void configureDrivebaseBindings() {
...
+    if (DriverStation.isDisabled()) {
+      s.drivebaseSubsystem
+          .applyRequest(() -> brake)
+          .ignoringDisable(true)
+          .withName("Brake")
+          .schedule();
```
jamesdooley4, line 158: Similar to above, I think this won't do anything until the robot becomes enabled since motors can't be actuated during disabled mode, so the wheels won't move to the X position until the robot is enabled (and assuming nothing else overrides the swerve command set here before the robot is enabled). That generally isn't a good thing, as when the robot is next enabled the wheels will immediately rotate out of X to whatever direction the robot wants to go, and that initial rotation out of X adds noise to the swerve odometry (and a delay in movement while the robot waits for the wheels to rotate before the drive motors start spinning.

src/main/java/frc/robot/Subsystems.java
jamesdooley4: Please re-enable

jamesdooley4 (changes requested):

KangarooKoala (changes requested):
Just did some discussion with Strummer and others, and we don't ever want to go into X, so we can just remove that code.

src/main/java/frc/robot/Controls.java
```diff
@@ -44,10 +44,12 @@ public class Controls {
   private BranchHeight branchHeight = null;
 
   // Swerve stuff
-  private double MaxSpeed =
+  public static double MaxSpeed =
       RobotType.getCurrent() == RobotType.BONK
```
jamesdooley4, line 48: This doesn't look like it needs to be `public` as I don't find anything outside this class that uses it. Also, it should be `final` since it never changes after initialization.

KangarooKoala (approved):

jamesdooley4 (dismissed):

KangarooKoala (approved):

PR #43, opened 2025-02-25
Spider+BigNorth Algae work

Skymouse888:
added algae preset poses on armpivot and elevator, added intake speed stuff on spinnyclaw, created rough algae control commands on superstructure. added test commands for armpivot alage stuff, haven't added test positions and speeds for spinnyclaw and superstructure in the controls class. i think that's everything?


iamawesomecat:
I was so confused by the title 😭

KangarooKoala:
Stuff to add for algae:
* [ ] Add ScoreMode enum in the util folder (similar to BranchHeight)
* [ ] Add `private ScoreMode scoreMode = ScoreMode.CORAL;` variable to Controls
* [ ] Add controls according to the OneNote controls plan to set the current score mode
* [ ] Use `Commands.select()` to check the current score mode to determine whether to use coral or algae mode

src/main/java/frc/robot/Controls.java
```diff
@@ -174,9 +180,38 @@ private void configureSuperStructureBindings() {
...
+    /*operatorController
+        .x()
+        .onTrue(
+            Commands.runOnce(() -> branchHeight = BranchHeight.ALGAE_LEVEL_THREE_FOUR).withName("algae level 3-4"));
+    operatorController
+        .b()
+        .onTrue(Commands.runOnce(() -> branchHeight = BranchHeight.ALGAE_LEVEL_TWO_THREE).withName("algae level 2-3"));
+    operatorController
+        .a()
+        .onTrue(Commands.runOnce(() -> branchHeight = BranchHeight.ALGAE_STOWED).withName("algae stow"));*/
```
jamesdooley4, line 197: Delete commented-out duplicate code

src/main/java/frc/robot/Controls.java
```diff
@@ -164,7 +167,10 @@ private void configureSuperStructureBindings() {
...
     operatorController
         .y()
-        .onTrue(Commands.runOnce(() -> branchHeight = BranchHeight.LEVEL_FOUR).withName("level 4"));
+        .onTrue(Commands.runOnce(() -> branchHeight = BranchHeight.LEVEL_FOUR).withName("level 4"))
+        .onTrue(
+            Commands.runOnce(() -> algaeIntakeHeight = AlgaeIntakeHeight.ALGAE_LEVEL_THREE_FOUR)
+                .withName("algae level 3-4"));
     operatorController
         .x()
         .onTrue(
```
jamesdooley4, line 176: Consider also adding algae level setting to the `x` and `b` buttons so the operator can hit either the 3 or 4 button to get the 3/4 algae and either the 1 or 2 button to get the 2/3 algae

src/main/java/frc/robot/subsystems/SuperStructure.java
```diff
@@ -93,4 +93,36 @@ public Command intake() {
...
+  }
+
+  public Command algaeStow() { // Big North + Spider collab on this one
+    return Commands.parallel(
+        elevator.setLevel(ElevatorSubsystem.ALGAE_STOWED),
+        armPivot.moveToPosition(ArmPivot.ALGAE_STOWED),
+        spinnyClaw.algaeIntakePower());
+  }
+
+  public Command algaeScore() { // Big North + Spider collab on this one
```
jamesdooley4, line 120: Would be better to call this `algaeDescore` as the command will actually be for descoring Algae. Also, I expect Algae will need both 'prep' and 'descore' commands for each level.

src/main/java/frc/robot/subsystems/SpinnyClaw.java
```diff
@@ -76,6 +86,13 @@ public Command holdExtakePower() {
         .withName("Hold extake power");
   }
 
+  /*
+  public Command algaeHoldIntakePower() { // tbh idk if it works
+    return startEnd(() -> motor.setVoltage(ALGAE_INTAKE_SPEED), () -> motor.stopMotor())
+        .withName("Algae hold intake power");
```
jamesdooley4, line 92: Testing showed this doesn't work, so I'd recommend deleting the method

src/main/java/frc/robot/Controls.java
```diff
@@ -41,7 +42,9 @@ public class Controls {
   private final Sensors sensors;
   private final SuperStructure superStructure;
 
-  private BranchHeight branchHeight = null;
+  private BranchHeight branchHeight = BranchHeight.LEVEL_FOUR;
+  private ScoringMode scoringMode = ScoringMode.CORAL;
```
jamesdooley4, line 46: There are no controls to change this, so the scoring mode is always `CORAL`. At least for testing there will need to be a way to change this to `ALGAE` and back to `CORAL`

jamesdooley4 (changes requested):

src/main/java/frc/robot/subsystems/SuperStructure.java
braelynandthefrogs: I believe algaeScore is referring to processor score, we can clarify the name

src/main/java/frc/robot/Controls.java
```diff
@@ -174,9 +180,37 @@ private void configureSuperStructureBindings() {
...
+    ;
+
+    /*operatorController
+    .a()
+    .onTrue(Commands.runOnce(() -> branchHeight = BranchHeight.ALGAE_STOWED).withName("algae stow"));*/
+
+    operatorController
+        .rightBumper()
+        .onTrue(superStructure.stow().withName("Stow"))
+        .onTrue(superStructure.algaeStow().withName("Algae Stow"));
```
KangarooKoala, line 196: What will happen when the operator right bumper is pressed is that first the normal `stow()` command will be scheduled and then immediately after wards the `algaeStow()` command will be scheduled, interrupting the normal `stow()`.

src/main/java/frc/robot/Controls.java
braelynandthefrogs: will talk to drive team

src/main/java/frc/robot/Controls.java
braelynandthefrogs: we decided to comment out algaestow and figure it out later if we ever score algae

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -234,6 +241,20 @@ public Command setLevel(double pos) {
...
+          m_motor.setControl(new StaticBrake());
+        });
+  }
+
+  public Command setCoastMode() {
+    return runOnce(
+        () -> {
+          m_motor.setControl(new CoastOut());
+        });
+  }
```
KangarooKoala, line 256:
```suggestion
  public Command holdCoastMode() {
    return startEnd(
          () -> {
            m_motor.setControl(new CoastOut());
          },
          () -> {
            m_motor.setControl(new StaticBrake());
          })
        .ignoringDisable(true)
        .withName("Hold elevator coast");
  }
```
By putting the coast and brake into the same command, we can guarantee we can't set coast without setting brake when we're done. This also moves more of the logic (ignoring disable and naming) into the command factory so that we won't need to duplicate that if we use this somewhere else (e.g., bind to a Trigger that represents a physical button on the robot).

src/main/java/frc/robot/Controls.java
```diff
@@ -272,6 +324,14 @@ private void configureElevatorBindings() {
...
                 .ignoringDisable(true)
                 .withName("Reset elevator zero"));
+    elevatorTestController
+        .leftBumper()
+        .whileTrue(
+            Commands.startEnd(
+                    () -> s.elevatorSubsystem.setCoastMode(),
+                    () -> s.elevatorSubsystem.setBrakeMode())
+                .ignoringDisable(true)
+                .withName("Coast to Brake Mode"));
```
KangarooKoala, line 334:
```suggestion
    elevatorTestController.leftBumper().whileTrue(s.elevatorSubsystem.holdCoastMode());
```
This goes hand in hand with the changes to the elevator factories. (Also, I double checked and this actually is a free button on elevatorTestController)

src/main/java/frc/robot/Controls.java
```diff
@@ -174,9 +179,42 @@ private void configureSuperStructureBindings() {
...
+    .a()
+    .onTrue(Commands.runOnce(() -> branchHeight = BranchHeight.ALGAE_STOWED).withName("algae stow"));*/
+    operatorController
+        .leftTrigger()
+        .onTrue(
+            Commands.runOnce(() -> scoringMode = ScoringMode.ALGAE).withName("Algae Scoring Mode"));
+    operatorController
+        .rightTrigger()
+        .onTrue(
+            Commands.runOnce(() -> scoringMode = ScoringMode.CORAL).withName("Coral Scoring Mode"));
```
KangarooKoala, line 198: Based on the OneNote, left and right trigger should both set coral scoring mode and left and right bumper should both set algae scoring mode.

src/main/java/frc/robot/subsystems/SuperStructure.java
```diff
@@ -22,75 +22,127 @@ public SuperStructure(
...
+            armPivot.moveToPosition(ArmPivot.CORAL_PRESET_L1),
             spinnyClaw.stop()),
         Commands.waitUntil(score),
-        elevator.setLevel(ElevatorSubsystem.LEVEL_TWO_POS),
+        elevator.setLevel(ElevatorSubsystem.CORAL_LEVEL_TWO_POS),
         spinnyClaw.holdExtakePower().withTimeout(0.15),
         stow());
   }
 
   public Command stow() {
```
KangarooKoala, line 75: We should rename superstructure stow and intake to coralStow and coralIntake to be clear about which one it does since we have both coral and algae versions. (Renaming the levelOne through levelFour presets to include coral would also be nice)

src/main/java/frc/robot/subsystems/SuperStructure.java
```diff
@@ -22,75 +22,127 @@ public SuperStructure(
...
         .withName("Intake");
   }
+
+  public Command algaeLevelThreeFourIntake() {
+    return Commands.sequence(
+        Commands.parallel(
+            spinnyClaw.algaeIntakePower(),
+            armPivot.moveToPosition(ArmPivot.ALGAE_REMOVE),
+            elevator.setLevel(ElevatorSubsystem.ALGAE_LEVEL_THREE_FOUR)),
+        algaeStow());
```
KangarooKoala, line 103: Theoretically speaking there should be a wait here and in `algaeLevelTwoThreeIntake()` so we don't immediately stow after reaching the intake position, but since we don't have algae intake mechanically right now it doesn't really matter. (You can just file this in your memory and resolve this conversation)

src/main/java/frc/robot/subsystems/SuperStructure.java
```diff
@@ -22,75 +22,127 @@ public SuperStructure(
...
+  }
+
+  public Command algaeLevelThreeFourFling() {
+    return Commands.sequence(
+        Commands.parallel(
+            spinnyClaw.algaeFlingPower(),
+            armPivot.moveToPosition(ArmPivot.ALGAE_REMOVE),
+            elevator.setLevel(ElevatorSubsystem.ALGAE_LEVEL_THREE_FOUR)),
+        algaeStow());
```
KangarooKoala, line 121: What does the fling process look like? I'm guessing we'll flinging out the algae, but I might be misunderstanding.
Rotating the arm would mean that this should look like the coral presets with a `Commands.parallel()` to go to the pre-score position, a `Commands.waitUntil()` to wait until the driver says to score, `armPivot.moveToPosition()` to rotate the arm and hit the algae, and then going back to stow. (The factory would have a `BooleanSupplier` parameter to say when to score, and we'd want different elevator heights (since instead of aligning to the height of the algae we align to below the algae) and separate `ALGAE_PRE_FLING` and `ALGAE_FLING` constants for the arm pivot angles.

KangarooKoala (commented):
Just remembered some things, oopsies.

KangarooKoala:
Before I forget- Drive team wants rightTrigger() to start the sequence and rightBumper() to finish the scoring sequence. (I think this just means moving the score button from leftBumper() to rightTrigger())

KangarooKoala:
New controls plan for now (checked with Tristan):
* [X] Driver left X, left Y, and right X for normal drive
* [X] Driver controller back for gyro reset
* [X] Driver A to force an intake sequence
* [ ] Driver right trigger to run the preset selected by the operator
* [X] Driver right bumper to finish the preset (after the robot is lined up)
* [ ] Driver POV down to select coral L1 / algae intake L2-L3
* [ ] Driver POV right to select coral L2
* [ ] Driver POV left to select coral L3
* [ ] Driver POV up to select coral L4 / algae intake L3-L4
* [X] Operator left Y for elevator manual control (press the stick to activate)
* [X] Operator right Y for arm manual control (press the stick to activate)
* [ ] Operator back to set elevator zero
* [ ] Operator start for climb
* [X] Operator left trigger to select coral mode
* [X] Operator left bumper to select algae mode
* [ ] Operator right bumper to put elevator into coast (only while held)
* [X] Operator A to select coral L1 / algae intake L2-L3
* [X] Operator B to select coral L2
* [X] Operator X to select coral L3
* [X] Operator Y to select coral L4 / algae intake L3-L4
* [ ] Operator POV/D-pad left to go to stow
* [X] Operator POV/D-pad right to swing arm straight forward (this is so that the arm is out of the way of the elevator coming down to zero)

src/main/java/frc/robot/Controls.java
KangarooKoala: Did some more thinking about the controls and right bumper will be used for putting the elevator into coast, so we can keep it as is.

KangarooKoala:
Operator controls aren't quite right- Coral scoring mode should be left bumper and coast mode should be right bumper.

Also, new request from me- Add a binding to the operator POV (either up or down should be available) to go to pre-intake (after the arm sensor PR gets merged and after merging in main).

jamesdooley4 (changes requested):

src/main/java/frc/robot/Controls.java
```diff
@@ -161,38 +161,92 @@ private void configureSuperStructureBindings() {
...
-        .onTrue(Commands.runOnce(() -> branchHeight = BranchHeight.ALGAE_STOWED).withName("algae stow"));*/
-    operatorController.rightBumper().onTrue(superStructure.stow().withName("Stow"));
+        .povLeft()
+        .onTrue(
+            Commands.select(
+                    Map.of(
+                        ScoringMode.CORAL,
+                        superStructure.coralStow().asProxy(),
+                        ScoringMode.ALGAE,
+                        superStructure.algaeStow().asProxy()),
```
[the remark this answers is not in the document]
KangarooKoala: For the select commands, we're using `.asProxy()` so that the command scheduler sees the specific selected command instead of just the select command as a whole, so the actual name shows up in the logs. I'm not completely sure I know what you're talking about for cancelling other commands using the super structure, since the composition is a select command (so the subcommands run for the same duration as the composition), not a sequential command group.

Now that I think about it, since we're proxying the commands, we could use switch statements (or switch expressions since we're on Java 17) inside `Commands.deferredProxy()`, which will hopefully make more sense to read.

**For the algae team-** Here's two examples of using `Commands.deferredProxy()` with different syntax:
```java
// Switch statements
Commands.deferredProxy(
  () -> {
    switch (scoringMode) {
      case CORAL -> {
        return superStructure.coralIntake();
      }
      case ALGAE -> {
        switch (algaeIntakeHeight) {
          case ALGAE_LEVEL_THREE_FOUR -> {
            return superStructure.levelThreeFourFling();
          }
          case ALGAE_LEVEL_TWO_THREE -> {
            return superStructure.levelTwoThreeFling();
          }
        }
      };
    }
  });

// Switch expressions
Commands.deferredProxy(
  () ->
    switch (scoringMode) {
      case CORAL -> superStructure.coralIntake();
      case ALGAE -> switch (algaeIntakeHeight) {
        case ALGAE_LEVEL_THREE_FOUR -> superStructure.levelThreeFourFling();
        case ALGAE_LEVEL_TWO_THREE -> superStructure.levelTwoThreeFling();
      };
    });
```

src/main/java/frc/robot/Controls.java
```diff
@@ -321,59 +371,59 @@ private void configureArmPivotBindings() {
...
-        .y()
-        .whileTrue(s.armPivotSubsystem.SysIDQuasistatic(Direction.kReverse));*/
-    armPivotSpinnyClawController
-        .leftStick()
+    operatorController // Untested :p
+        .rightStick()
+        .whileTrue(
+            s.armPivotSubsystem
+                .startMovingVoltage(() -> Volts.of(3 * operatorController.getRightY()))
+                .withName("Arm Manual Control"));
```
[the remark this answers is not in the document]
KangarooKoala: This has been tested 3/3 (but we should remove the comment and probably note that forward on joystick is forward/down on arm).

src/main/java/frc/robot/Controls.java
```diff
@@ -253,6 +371,7 @@ private void configureElevatorBindings() {
                         .withTimeout(0.3))
                 .ignoringDisable(true)
                 .withName("Reset elevator zero"));
+    operatorController.rightBumper().whileTrue(s.elevatorSubsystem.holdCoastMode());
```
KangarooKoala, line 374: Note from testing- This binding worked, but it appears that most of the resistance is from friction (not brake mode on the motors), so this doesn't make it much easier to move (and it doesn't naturally fall past the first stage).

KangarooKoala (commented):
Some cleanup we can do and some notes from 3/3 testing.

src/main/java/frc/robot/Controls.java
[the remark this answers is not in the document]
Skymouse888: have used a better way of proxy in the 3 instances in which i found it

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -252,6 +261,18 @@ public Command setLevel(double pos) {
         .withName("setLevel" + pos);
   }
 
+  public Command holdCoastMode() {
+    return startEnd(
+            () -> {
+              m_motor.setControl(new CoastOut());
```
KangarooKoala, line 267: We should test a) if this sets both motors and b) if this works when disabled.

src/main/java/frc/robot/subsystems/SuperStructure.java
```diff
@@ -20,90 +20,177 @@ public SuperStructure(
...
             spinnyClaw.stop()),
-        armPivot.moveToPosition(ArmPivot.PRESET_L2_L3),
+        armPivot.moveToPosition(ArmPivot.CORAL_PRESET_L2_L3),
         Commands.waitUntil(score),
-        elevator.setLevel(ElevatorSubsystem.LEVEL_THREE_POS),
-        spinnyClaw.holdExtakePower().withTimeout(0.15),
-        preIntake());
+        elevator.setLevel(ElevatorSubsystem.CORAL_LEVEL_THREE_POS),
+        spinnyClaw.coralHoldExtakePower().withTimeout(0.15),
+        preIntake().withName("Coral Level 3"));
```
KangarooKoala, line 49:
```suggestion
        preIntake()).withName("Coral Level 3");
```
This makes the `.withName()` apply to the entire command, not just the `preIntake()` command.

KangarooKoala (commented):
I'm really tempted to just merge this into main so that other PRs can incorporate these changes and fix the rest later.

Also, it would be nice to have a button to go to pre-intake.

Spinny claw stalling also needs to be investigated soon.

src/main/java/frc/robot/Controls.java
```diff
@@ -140,7 +142,10 @@ private void configureSuperStructureBindings() {
     // operator start button used for climb - bound in climb bindings
     operatorController
         .y()
-        .onTrue(Commands.runOnce(() -> branchHeight = BranchHeight.LEVEL_FOUR).withName("level 4"));
+        .onTrue(Commands.runOnce(() -> branchHeight = BranchHeight.LEVEL_FOUR).withName("level 4"))
+        .onTrue(
+            Commands.runOnce(() -> algaeIntakeHeight = AlgaeIntakeHeight.ALGAE_LEVEL_THREE_FOUR)
+                .withName("algae level 3-4"));
```
jamesdooley4, line 148: Consider this to avoid creating and running a second command simply to set a variable, and updating the similar ones below for the `a`, `povUp`, and `povDown` buttons.
```suggestion
        .onTrue(Commands.runOnce(() -> {
            branchHeight = BranchHeight.LEVEL_FOUR;
            algaeIntakeHeight = AlgaeIntakeHeight.ALGAE_LEVEL_THREE_FOUR;
        }).withName("coral level4, algae level 3-4"));
```

src/main/java/frc/robot/Subsystems.java
```diff
@@ -20,11 +20,11 @@ public static class SubsystemConstants {
     // <SUBSYSTEM>_ENABLED constants go here
 
     public static final boolean DRIVEBASE_ENABLED = true;
-    public static final boolean VISION_ENABLED = true;
+    public static final boolean VISION_ENABLED = false;
     public static final boolean ELEVATOR_ENABLED = true;
     public static final boolean ARMPIVOT_ENABLED = true;
     public static final boolean SPINNYCLAW_ENABLED = true;
-    public static final boolean CLIMBPIVOT_ENABLED = true;
+    public static final boolean CLIMBPIVOT_ENABLED = false;
```
jamesdooley4, line 27: Remember to re-enable these once they're working again.

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
jamesdooley4: On Crane it worked while disabled to toggle the motor between brake & coast mode, though those were SparkMax controllers.

jamesdooley4: May also need to do `m_motor.stopMotor()` to abort any previous position control request, as that would continue to work even if the motor was in coast mode.

jamesdooley4 (approved):

KangarooKoala:
Merging into main so other PRs can pull in the code.

PR #47, opened 2025-02-26
Arm sensor work

Skymouse888:


src/main/java/frc/robot/subsystems/SuperStructure.java
```diff
@@ -79,13 +80,24 @@ public Command stow() {
         spinnyClaw.stop());
   }
 
+  public Command preIntake() {
+    return Commands.sequence(
+        stow(),
```
jamesdooley4, line 85: Why does it need to move to the stowed position first (arm angled up) and then move to the pre-intake position (arm pointing straight down)?

KangarooKoala: It probably doesn't, my mistake from when I was telling him how to write the sequence. (I originally had `stow()` in the beginning so that it would act the same as the old intake sequence except the pre-intake happened without needing to press the button, but we can just directly go into the pre-intake position)

src/main/java/frc/robot/sensors/ArmSensor.java
```diff
@@ -1,27 +1,32 @@
...
 
 public class ArmSensor {
 
   private final LaserCan mainSensor;
-  // Sensor reads in meters, is converted to milimeters in code?
-  private static final double TROUGH_LOWER_LIMIT = 250; // untested guess value (in millimeters)
-  private static final double TROUGH_UPPER_LIMIT = 350; // untested guess value
-  private static final double CLAW_LOWER_LIMIT = 30; // untested guess value
-  private static final double CLAW_UPPER_LIMIT = 130; // untested guess value
+  // VALUES ARE IN METERES
```
jamesdooley4, line 17: Typo: Meters

src/main/java/frc/robot/sensors/ArmSensor.java
```diff
@@ -46,23 +51,24 @@ public Distance getSensorDistance() {
   public Trigger inTrough() {
     return new Trigger(
         () -> {
-          if (getSensorDistance().in(Millimeters) > TROUGH_LOWER_LIMIT
-              && getSensorDistance().in(Millimeters) < TROUGH_UPPER_LIMIT) {
+          if (getSensorDistance().in(Meters) > TROUGH_LOWER_LIMIT
+              && getSensorDistance().in(Meters) < TROUGH_UPPER_LIMIT) {
```
jamesdooley4, line 55: Would be better to read the distance once above the `if()` statements and then compare against that, to make sure the value doesn't change between the two checks (theoretically possible).

jamesdooley4: Ditto from below

jamesdooley4 (dismissed):

src/main/java/frc/robot/sensors/ArmSensor.java
```diff
@@ -46,23 +51,24 @@ public Distance getSensorDistance() {
...
-    };
+  public Trigger inClaw() {
+    return new Trigger(
+        () -> {
+          if (getSensorDistance().in(Meters) > CLAW_LOWER_LIMIT
+              && getSensorDistance().in(Meters) < CLAW_UPPER_LIMIT) {
+            return true;
+          } else {
+            return false;
+          }
```
markpete, line 71: You don't need to do quite this much.  Do similar to what you did for InPreIntakePostion().

`return new Trigger(
    () -> {
      var distance = getSensorDistance().in(Meters);
       return (distance > CLAW_LOWER_LIMT && distance < CLAW_UPPER_LIMIT);
   )
);`

markpete: Do similar above.  No need to return true/false;

KangarooKoala (approved):
Worked in testing, if there's code stuff to fix we'll do that later.

PR #48, opened 2025-02-28
Autos

SnappleRamen:
Autos branch ready to be put in main


src/main/java/frc/robot/util/AutoLogic.java
```diff
@@ -0,0 +1,248 @@
...
+              return alliance.get() == DriverStation.Alliance.Red;
+            }
+            return false;
+          },
+          s.drivebaseSubsystem // Reference to this subsystem to set requirements
+          );
+
+    } catch (IOException | ParseException e) {
+      // TODO Auto-generated catch block
+      e.printStackTrace();
```
jamesdooley4, line 82: Would be good to also call `DriverStation.reportError` to more visibly surface the error

src/main/java/frc/robot/util/AutoLogic.java
```diff
@@ -0,0 +1,248 @@
...
+        .withName("raiseElevatorandWAIT");
+  }
+
+  public static Command autoAlignmentCommand() {
+    return AutoAlign.autoAlign(s.drivebaseSubsystem).withName("AutoAlign");
+  }
+
+  public static void registerCommand() {
+
+    if (RobotState.isDisabled()) {
```
jamesdooley4, line 229: It is possible for the robot to be enabled while running the code constructors. That will happen if the rio reboots (browns out) mid-match as the WPILib code can get the 'enabled' signal before the various class / Java initializers run.
So, it's not safe to run this only if the robot is initialized - some other method should be used to ensure these are only registered once (if that's the intention) or in certain states.

jamesdooley4 (changes requested):

iamawesomecat:
How many of these comments were resolved in the recent commits?

src/main/deploy/pathplanner/settings.json
```diff
@@ -0,0 +1,39 @@
+{
+  "robotWidth": 0.743,
+  "robotLength": 0.8509,
+  "holonomicMode": true,
+  "pathFolders": [
+    "New Folder"
```
markpete, line 6: Can we have more descriptive names other than "New Folder" and "New New Folder"?

src/main/java/frc/robot/Robot.java
```diff
@@ -96,7 +107,21 @@ public void disabledPeriodic() {}
   public void disabledExit() {}
 
   @Override
-  public void autonomousInit() {}
+  public void autonomousInit() {
+
+    // Checks if FMS is attatched and enables joystick warning if true
```
markpete, line 112: "attached"  (spelling)

src/main/java/frc/robot/generated/TestBaseTunerConstants.java
```diff
@@ -0,0 +1,314 @@
...
+  private static final SteerMotorArrangement kSteerMotorType =
+      SteerMotorArrangement.TalonFX_Integrated;
+
+  // The remote sensor feedback type to use for the steer motors;
+  // When not Pro-licensed, Fused*/Sync* automatically fall back to Remote*
+  private static final SteerFeedbackType kSteerFeedbackType = SteerFeedbackType.FusedCANcoder;
+
+  // The stator current at which the wheels start to slip;
+  // This needs to be tuned to your individual robot
+  private static final Current kSlipCurrent = Amps.of(120.0);
```
markpete, line 58: Is 120 Amps correct here?  I thought James said our breakers are at 40 Amps.

src/main/java/frc/robot/generated/TestBaseTunerConstants.java
```diff
@@ -0,0 +1,314 @@
...
+  // Some configs will be overwritten; check the `with*InitialConfigs()` API documentation.
+  private static final TalonFXConfiguration driveInitialConfigs = new TalonFXConfiguration();
+  private static final TalonFXConfiguration steerInitialConfigs =
+      new TalonFXConfiguration()
+          .withCurrentLimits(
+              new CurrentLimitsConfigs()
+                  // Swerve azimuth does not require much torque output, so we can set a relatively
+                  // low
+                  // stator current limit to help avoid brownouts without impacting performance.
+                  .withStatorCurrentLimit(Amps.of(60))
```
markpete, line 70: Ditto above.  Dbl-check this.

src/main/java/frc/robot/subsystems/ArmPivot.java
```diff
@@ -31,13 +31,13 @@ public class ArmPivot extends SubsystemBase {
   private final double ARMPIVOT_KV = 0.69;
   private final double ARMPIVOT_KG = 0.18;
   private final double ARMPIVOT_KA = 0.0;
-  public static final double PRESET_L1 = -1.0 / 16; // This is at position perpendicular to elevator
+  public static final double PRESET_L1 = -1.0 / 16;
```
markpete, line 34: If the comment is incorrect, can you add a comment that is correct?

src/main/java/frc/robot/Controls.java
```diff
@@ -44,14 +47,15 @@ public class Controls {
...
           ? BonkTunerConstants.kSpeedAt12Volts.in(MetersPerSecond)
-          : CompTunerConstants.kSpeedAt12Volts.in(MetersPerSecond);
-  private final double MAX_ACCELERATION = 50;
-  private final double MAX_ROTATION_ACCELERATION = 50;
+          : RobotType.getCurrent() == RobotType.TESTBASE
+              ? TestBaseTunerConstants.kSpeedAt12Volts.in(MetersPerSecond)
+              : CompTunerConstants.kSpeedAt12Volts.in(MetersPerSecond);
   // kSpeedAt12Volts desired top speed
-  private double MaxAngularRate =
+  public static double MaxAngularRate =
```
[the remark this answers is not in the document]
jamesdooley4: And add `final` back to `MaxSpeed` above.

src/main/java/frc/robot/Controls.java
```diff
@@ -340,4 +340,23 @@ private void configureElevatorLEDBindings() {
           s.elevatorLEDSubsystem.colorSet(50, 0, 0).withName("LED red").ignoringDisable(false));
     }
   }
+
+  private void configureAlignBindings() {
+    if (s.drivebaseSubsystem == null) {
+      return;
+    }
+    driverController.leftBumper().onTrue(AutoAlign.autoAlign(s.drivebaseSubsystem));
```
jamesdooley4, line 348: Needs to go in after #43 that moves the existing `leftBumper` control (see line 165 in this version of the file) to a different button.

src/main/java/frc/robot/subsystems/SuperStructure.java
```diff
@@ -22,6 +22,7 @@ public SuperStructure(
 
   public Command levelFour(BooleanSupplier score) {
     return Commands.sequence(
+        Commands.print("Running L4"),
```
jamesdooley4, line 25: #43 adds names for these, which removes the need for debug printing. So please delete this.

src/main/java/frc/robot/subsystems/SuperStructure.java
```diff
@@ -104,6 +105,6 @@ public Command intake() {
             spinnyClaw.stop(),
             elevator.setLevel(ElevatorSubsystem.PRE_INTAKE),
             stow())
-        .withName("Intake");
+        .withName("intake");
```
jamesdooley4, line 108: I'd recommend reverting this to avoid a future merge conflict since #43 changes this name to something else,

src/main/java/frc/robot/util/AutoLogic.java
```diff
@@ -0,0 +1,235 @@
...
+    autoPicker.addOption("TESTING Low 3 Piece", "TEST Low3Piece");
+    autoPicker.addOption("ALIGNMENT TEST", "ALIGN TEST");
+    autoPicker.addOption("DRIVE TEST", "DRIVE1FT");
+  }
+
+  public static void addTestedAutos() {}
+
+  public static Command raiseElevator() {
+    if (r.superStructure != null) {
+      return r.superStructure.levelFour(() -> true).asProxy();
```
jamesdooley4, line 198: `.asProxy()` should only be needed if this is inside a composite command (parallel, sequential, etc.). Since that's unknowable here, I'd recommend removing it here & adding it in whatever composite command is calling this (and only if there are no other options to avoid its use).

src/main/java/frc/robot/util/AutoLogic.java
```diff
@@ -0,0 +1,235 @@
...
+    return Commands.none().withName("raiseElevator");
+  }
+
+  public static Command autoAlignmentCommand() {
+    return AutoAlign.autoAlign(s.drivebaseSubsystem).withName("autoAlign");
+  }
+
+  public static Command intakeCommand() {
+    if (r.superStructure != null) {
+      return r.superStructure.intake().withName("intake").asProxy();
```
jamesdooley4, line 209: Same comment as for `raiseElevator`'s use of `asProxy()`

jamesdooley4 (changes requested):

KangarooKoala (approved):
Tested on 3/7 before pit load-in. One piece works and the logic for three piece works, but odometry gets pretty bad when going to the reef for the second and third times.

I just realized we'll need to double-check the alliance handling as well- It should work out of the box, but it'd be good to double check on a practice field first.

Auto align wasn't tested today, but drive team ran some version of auto align 3/5 and it seemed to mostly work.

KangarooKoala:
There's a decent amount of cleanup to do afterwards, but the DEFAULT PATH and Mid3Piece autos work and we need those in for competition.

PR #50, opened 2025-03-01
Climb encoder

RobototesProgrammers:
adds climb encoder functionality, holds funnel down.

### DIFF
--- a/src/main/java/frc/robot/subsystems/ClimbPivot.java
+++ b/src/main/java/frc/robot/subsystems/ClimbPivot.java
@@ -35,7 +35,7 @@ public class ClimbPivot extends SubsystemBase {
   private final double CLIMB_IN_SPEED = 0.2;
   private final double CLIMB_OUT_SPEED = -0.2;
   private final double BOOLEAN_TOLERANCE = 2;
-  //relative to eachother, likely not accurately zero'ed when obtained.
+  // relative to eachother, likely not accurately zero'ed when obtained.x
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/subsystems/ClimbPivot.java
```diff
@@ -31,23 +35,37 @@ public class ClimbPivot extends SubsystemBase {
   private final double CLIMB_IN_SPEED = 0.2;
   private final double CLIMB_OUT_SPEED = -0.2;
   private final double BOOLEAN_TOLERANCE = 2;
+  private static final double MIN_ROTOR_POSITION = -50.45;
+  private static final double MAX_ROTOR_POSITION = 14.456;
```
jamesdooley4, line 39: FYI that these are relative to where the motor was when it started, and I suspect they were read without doing a `zeroClimb` first since one of them is negative. They're fine to use as relative to each other (as is done below) but shouldn't be used for any absolute references. I'd suggest adding a comment to that effect so future programmers don't mistakenly try to use them for comparisons with the read motor rotor position.

jamesdooley4 (approved):

src/main/java/frc/robot/subsystems/ClimbPivot.java
RobototesProgrammers: I will make note of that.

KangarooKoala (approved):

PR #51, opened 2025-03-02
vision merge into main

RobototesProgrammers:
it has main bot dimensions


src/main/java/frc/robot/subsystems/VisionSubsystem.java
```diff
@@ -152,6 +178,7 @@ private void process(PhotonPipelineResult result, PhotonPoseEstimator estimator)
         lastFieldPose = FieldPose;
         rawVisionFieldObject.setPose(lastFieldPose);
       }
+      this.Distance = Distance;
```
jamesdooley4, line 181: Is this intentionally updated even if the pose time is older than the most recent valid pose?

jamesdooley4 (approved):

src/main/java/frc/robot/subsystems/VisionSubsystem.java
koolpoolo: the only reason I did that is that I have two Distance variable and I don't know to fix it otherwise but we should fix it if it can do that

koolpoolo: should it be in the if (RawTimestampSeconds > lastRawTimestampSeconds) statment at the end

KangarooKoala: Yeah, I think that's what James is talking about.

koolpoolo: ok I will put it in there

koolpoolo: its in there now

markpete (approved):

jamesdooley4 (approved):

PR #54, opened 2025-03-07
Elevator leds

sabellad:
updated LEDs updated switched to wpilib LED pattern


KangarooKoala (approved):
Tested on 3/6. (There's some bugs which are fixed in #55)

PR #60, opened 2025-03-09
Changes from WASAM

KangarooKoala:
Draft because the event is still in progress. Opening now for visibility.


RobototesProgrammers:
Vision hasn't been getting many detections, so we needed to fudge the auto location in 347d6b5. Long term, we should fix vision not getting any results. (We should also double-check the PathPlanner settings and positions are correct, since visually in PathPlanner it doesn't go all the way to the coral station)

KangarooKoala:
Event's finished now.

src/main/java/frc/robot/subsystems/VisionSubsystem.java
```diff
@@ -178,6 +179,12 @@ private void process(PhotonPipelineResult result, PhotonPoseEstimator estimator)
     if (estimatedPose.isPresent()) {
       var TimestampSeconds = estimatedPose.get().timestampSeconds;
       var FieldPose3d = estimatedPose.get().estimatedPose;
+      RawfieldPose3dEntry.set(FieldPose3d);
+      if (!MathUtil.isNear(0, FieldPose3d.getZ(), 0.10)
+          || !MathUtil.isNear(0, FieldPose3d.getRotation().getX(), Units.degreesToRadians(3))
+          || MathUtil.isNear(0, FieldPose3d.getRotation().getY(), Units.degreesToRadians(3))) {
```
jamesdooley4, line 185: This is worth a comment explaining what this filtering is doing. I'm also not sure why a X rotation larger than 3 degrees is a good thing, but a Y rotation larger than 3 degrees is bad.

jamesdooley4 (approved):

src/main/java/frc/robot/subsystems/VisionSubsystem.java
KangarooKoala: Nice catch! That's a bug- I guess we forgot the `!` on the last one.

PR #62, opened 2025-03-15
Superstructure + spinny claw

Skymouse888:


src/main/java/frc/robot/sensors/ArmSensor.java
```diff
@@ -58,11 +59,15 @@ public Trigger inTrough() {
...
+  }
+
+  public boolean booleanInClaw() {
+    double distance = getSensorDistance().in(Meters);
+    return distance > CLAW_LOWER_LIMIT && distance < CLAW_UPPER_LIMIT;
+  }
+
+  public BooleanSupplier booleanSupplierInClaw() {
+    return () -> booleanInClaw();
   }
```
KangarooKoala, line 72: I love the independent problem solving! As a note, though, every `Trigger` is a `BooleanSupplier`, so we should use `inClaw()` instead of `booleanSupplierInClaw()`. (This also lets us use the debouncing in `inClaw()` to handle erratic inputs)

src/main/java/frc/robot/subsystems/SpinnyClaw.java
```diff
@@ -73,18 +76,41 @@ public void configMotors() {
   }
 
   private Command setPower(double pow) {
-    return runOnce(() -> motor.setVoltage(pow));
+    if (armSensor != null) {
+      return runOnce(
+          () -> {
+            if (armSensor.booleanInClaw()) {
```
KangarooKoala, line 82: We should also check that `pow` is negative, since if `pow` is positive that will spin the wheels out, which is fine.

src/main/java/frc/robot/sensors/ArmSensor.java
braelynandthefrogs: i just checked that and it worked, thanks. i could've sworn it didn't work before but knowing me i was probably just going crazy...

src/main/java/frc/robot/subsystems/SpinnyClaw.java
braelynandthefrogs: should we only allow it to be positive or should it also include zero?

braelynandthefrogs: nevermind

KangarooKoala (approved):
Not 100% tested or done, but we're merging into main in preparation for WAAHS.

PR #63, opened 2025-03-15
Testing 3/15

RobototesProgrammers:
Updated visions positions. Tested climb changes from Thursday. Merged in autos 2. Autos 2 changes: added shuffleboard organization for sorting autos, use autopath class to sort autos, changed all autos and paths for ones wanted for comp, added intake checks to know when to use auto intake and manual intake commands, updated some auto align positions. Added command interrupter check.


KangarooKoala (approved):
Approving so we can merge for now, but we'll still want to look for anything that should be improved for the upcoming meetings.
(Also wow, 96 changed files that are just from the auto cleanup)

PR #91, opened 2025-03-25
post WAAHS climb fixes/improvements

### LINKED ISSUE #85: Fix climb default command
Based on WAAHS E4 and E6 logs, it runs, but reportedly during testing "only ran once".

We should also add a name to the command.

### LINKED ISSUE #89: Fix climb inTolerance logging
Right now the value is updated once at the start and never updated again.

### LINKED ISSUE #90: Make separate climb speeds for different movement stages
This way, we can run faster on the way out and slower on the way in.

iamawesomecat:
Fixes #89 
Fixes #90 
Fixes #85

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -489,11 +489,9 @@ public class Controls {
       return;
     }
 
-    Command setClimbLEDs;
     if (s.elevatorLEDSubsystem != null) {
-      setClimbLEDs = s.elevatorLEDSubsystem.pulse(0, 0, 255, "Blue - Climb Extended");
-    } else {
-      setClimbLEDs = Commands.none();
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/Controls.java
```diff
@@ -496,7 +496,7 @@ private void configureClimbPivotBindings() {
       setClimbLEDs = Commands.none();
     }
 
-    s.climbPivotSubsystem.setDefaultCommand(s.climbPivotSubsystem.advanceClimbCheck());
+    s.climbPivotSubsystem.setDefaultCommand(s.climbPivotSubsystem.advanceClimbCheck().withName("Check Climb"));
```
markpete, line 499: Make the comment "Advance Climb Check", so that the log comment matches the API making it easier to map to.

KangarooKoala (approved):
Tested 3/25.

PR #100, opened 2025-03-25
apriltag filtering + standard deviation fixes

koolpoolo:


src/main/java/frc/robot/subsystems/VisionSubsystem.java
```diff
@@ -218,4 +221,14 @@ public double getLastRawTimestampSeconds() {
   public double getDistanceToTarget() {
     return (double) Math.round(Distance * 10) / 10;
   }
+
+  // configured for 2025 reefscape
+  private static boolean BadAprilTagDetector(PhotonPipelineResult r) {
+    for (var t : r.getTargets()) {
+      if (0 < t.fiducialId && t.fiducialId < 6 || 11 < t.fiducialId && t.fiducialId < 17) {
```
KangarooKoala, line 228: Thinking more about this, it might be clearer to specify the valid range of IDs (the reef IDs) rather than filtering out ones that are on the field but not reef tags.

koolpoolo: what do you mean?

koolpoolo: like do you want me to change it so that if the tag is in a vaild range it doesn't reject it but if its anything else then it rejects it

KangarooKoala: Do something like this:
```java
boolean isRed = ...
boolean isBlue = ...

boolean isRedReef = ...
boolean isBlueReef = ...
boolean isValid = isBlueReef && !isRed || isRedReef && !isBlue;
if (!isValid) {
  return true;
}
```
Splitting it into multiple variables also implicitly documents what each part of the condition does.

KangarooKoala (approved):
Tested tag fitlering 3/29.

PR #101, opened 2025-03-26
Elevator LEDs2

sabellad:
ok ok ok


src/main/java/frc/robot/Controls.java
```diff
@@ -642,20 +644,38 @@ private void configureElevatorLEDBindings() {
...
     }
     RobotModeTriggers.autonomous()
         .whileTrue(s.elevatorLEDSubsystem.animate(LEDPattern.rainbow(255, 255), "Auto Rainbow"));
+    s.elevatorLEDSubsystem
+        .thirtySecondsLeft()
+        .onTrue(
+            Commands.sequence(
+                s.elevatorLEDSubsystem.blinky(255, 0, 0, "red half blink - 30 sec remaining"),
+                s.elevatorLEDSubsystem.blinky(
+                    255, 255, 0, "Yellow half blink - 30 sec remaining")));
```
KangarooKoala, line 671: Right now, this won't work since each `blinky()` command will repeat on-off for its color forever. Instead, we can do a red-yellow blink by running solid red (using `colorSet()`) with a timeout, then running solid yellow (using `colorSet()`) with a timeout, all in a repeating sequence. (This means that we can remove the `blinky()` factory method)

src/main/java/frc/robot/sensors/ElevatorLight.java
```diff
@@ -103,4 +118,12 @@ public LEDPattern greenProgress(DoubleSupplier progress) {
     LEDPattern heightDisplay = base.mask(mask);
     return heightDisplay;
   }
+
+  public Trigger thirtySecondsLeft() {
```
KangarooKoala, line 122: This isn't inherently tied to the LEDs and should instead go in `Controls.java()`.

src/main/java/frc/robot/sensors/ElevatorLight.java
```diff
@@ -23,6 +25,7 @@
 public class ElevatorLight extends SubsystemBase {
 
   private CANdle candle;
+  private Timer timer = new Timer();
```
KangarooKoala, line 28: This timer is never started, so it never measures the specified time as having elapsed. The timer (which should probably be moved to `Controls.java`) should be started when we enter teleop less than five seconds after leaving auto and stopped and reset when we leave teleop. (`RobotModeTriggers.autonomous().debounce(5).and(RobotModeTriggers.teleop())` produces a trigger that is true while in teleop less than 5 seconds after being in auto for more than 5 seconds.)

KangarooKoala (approved):
Tested 4/8.

PR #106, opened 2025-03-28
Climb work 3/27

### LINKED ISSUE #104: Update climb logic
Previously, the ratchet required the motors to only run in one direction, making us rely on the winch string unspooling and respooling to flip the direction of motion of the pivot while the motors kept rotating in the same direction. This change in pivot rotation direction compared to motor rotation direction made the logic harder to reason about.
However, the ratchet is being replaced with a latch, so we can **run the motors in both directions**. This allows us to do an **active hold** and to **recover from any accidental rising** without running all the way out and back in (which would cause the climb latch to deploy).

We will also need to **verify the number of motors** and determine if **preset positions** need to be adjusted.

iamawesomecat:
various small climb fixes, made it work with new funnel hopefully maybe

fixes #104

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -519,9 +519,7 @@ public class Controls {
             s.climbPivotSubsystem
                 .moveClimbManual(
                     () ->
-                        0.15
-                            * -MathUtil.applyDeadband(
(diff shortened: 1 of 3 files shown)


KangarooKoala (approved):
Tested manual control and positions on 3/27. (Preset movement logic was not tested, however)

Merging for now so we can setup the drive practice branch.

PR #110, opened 2025-03-28
Various climb fixes 3/28

### LINKED ISSUE #108: Post-#106 Climb fixes
Missed a few things earlier:
- [x]  Run the motor in either direction depending on the direction funnel needs to move. Right now it only bases the direction and speed based on the target position, regardless of where the funnel is right now.

- [x] Prevent climb from running immediately upon reenable. We should tell the motors to stop when we enter disabled motor to clear any previous setpoint, like how [elevator does it](https://github.com/robototes/Reefscape2025/blob/ad2de345cd312866bdf0e494c65b643e83374e30/src/main/java/frc/robot/Controls.java#L368).

- [x] Delete commented out code in `periodic()`.
- [x] Update inTolerance and moveComplete in `periodic()`, so that updates always happen and all that the default command does is check the variables and set the speed accordingly.

We'll also need to see whether the presets work.

iamawesomecat:
fixes #108

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -535,9 +535,7 @@ public class Controls {
     connected(climbTestController)
         .and(climbTestController.start())
         .onTrue(s.climbPivotSubsystem.advanceClimbTarget());
-    // operatorController
-    //     .start()
-    //     .onTrue(s.climbPivotSubsystem.advanceClimbTarget());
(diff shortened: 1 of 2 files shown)


KangarooKoala (commented):
Nice! Now that you reminded me, could you also remove the commented out `moveClimbMotor` closer to the top of the file?

It would also be nice to rename `moveCompleteTrue()` to something like `haltMovement()` since it doesn't just set moveComplete to be true.

KangarooKoala (approved):
Tested (including presets!) 3/29.

PR #113, opened 2025-03-29
Ground intake

Polyehteretherketone:
ground intake setup


KangarooKoala (approved):
Tested running the ground spinny wheels and moving ground arm to intake position on 3/29.

PR #114, opened 2025-03-29
Autos dcmp

RobototesProgrammers:


KangarooKoala (approved):
Confirmed OSWRSF_D-C as a working 1.8\* piece auto 3/29.

\*1.8 piece means that we score one, intake another, and get into scoring position but don't actually finish the scoring sequence, so we don't get auto points for the coral but are ready for an almost immediately teleop score.

PR #121, opened 2025-04-02
fixed LED null...  issue #119 :D!!!

### LINKED ISSUE #119: Handle disabling the LED subsystem
Right now, disabling the LED subsystem causes a crash because we use LED animations in `Controls.java` without first checking if the LED subsystem is null. We can check if the LED subsystem is null and use `Commands.none()` instead of the LED animation if so.

I believe @Roxie222 has started working on this, but @sabellad should also be involved as the primary LED person.

RobototesProgrammers:
Closes #119


KangarooKoala (approved):
Tested 4/1.

PR #123, opened 2025-04-05
Three piece auto for the right side of the feild to use during eliminations of PNCMP

RobototesProgrammers:


KangarooKoala:
Superseded by #124.

PR #126, opened 2025-04-05
Barge align

KangarooKoala:
We need to merge main into this branch, since this was branched off of PNCMP which was then merged separately.


RobototesProgrammers:
Logic tested 4/7, don't know about the positions though.

src/main/java/frc/robot/Controls.java
```diff
@@ -349,7 +350,16 @@ private void configureSuperStructureBindings() {
...
+                                    BargeAlign.driveToBlackLine(s.drivebaseSubsystem),
+                                    BargeAlign.driveToBarge(s.drivebaseSubsystem)
+                                        .until(
+                                            () ->
+                                                BargeAlign.atScoringXPosition(s.drivebaseSubsystem))
+                                        .withDeadline(
+                                            superStructure.algaeNetScore(
+                                                () ->
+                                                    BargeAlign.atScoringXPosition(
+                                                        s.drivebaseSubsystem))),
```
jamesdooley4, line 362: This seems odd, as `atScoringXPosition` being true will interrupt `driveToBarge` and should also then cancel `algaeNetScore` since it's all part of the whole command composition group. I'd expect the algae to not fully leave the grabber since `algaeNetScore` won't run `algaeHoldExtakePower` for the full 0.7s.

jamesdooley4: Though maybe `withDeadline` overrides the `until()` interrupt and extends the command group until `algaeNetScore` ends even though `driveToBarge().until()` would have already ended?

src/main/java/frc/robot/subsystems/auto/BargeAlign.java
```diff
@@ -0,0 +1,113 @@
...
+
+  private CommandSwerveDrivetrain drive;
+  private static final double fieldLength = 17.548; // Welded field
+  private static final double blueBlacklineX = 7.09;
+  private static final double redBlacklineX = fieldLength - blueBlacklineX;
+  private static final double blueBargeLineX = 7.67;
+  private static final double redBargeLineX = fieldLength - blueBargeLineX;
+
+  private final SwerveRequest.FieldCentric blackLineDriveRequest =
+      new SwerveRequest.FieldCentric() // Add a 10% deadband
```
jamesdooley4, line 26: ? What does the `Add a 10% deadband` comment refer to?

src/main/java/frc/robot/subsystems/auto/BargeAlign.java
```diff
@@ -0,0 +1,113 @@
...
+    double targetX = onRedSide ? redBlacklineX : blueBlacklineX;
+    Rotation2d targetAngle = onRedSide ? Rotation2d.k180deg : Rotation2d.kZero;
+    pidX.setSetpoint(targetX);
+    pidRotate.setSetpoint(targetAngle.getRadians());
+  }
+
+  @Override
+  public void execute() {
+    Pose2d currentPose = drive.getState().Pose;
+    // Calculate the power for X direction and clamp it between -1 and 1
```
jamesdooley4, line 84: ? Comment says 'clamp it between -1 and 1' but the code appears to clamp between -2 and 2

src/main/java/frc/robot/Controls.java
```diff
@@ -349,7 +350,16 @@ private void configureSuperStructureBindings() {
                           switch (scoringMode) {
                             case CORAL -> getCoralBranchHeightCommand();
                             case ALGAE -> Commands.sequence(
-                                    superStructure.algaeNetScore(driverController.rightBumper()),
+                                    BargeAlign.driveToBlackLine(s.drivebaseSubsystem),
+                                    BargeAlign.driveToBarge(s.drivebaseSubsystem)
```
jamesdooley4, line 354: It seems inefficient and potentially risky to stop abruptly between these two commands and then start up again (as `BargeAlign.end()` in `driveToBlackLine` stops all motion). Ideally `driveToBlackLine` ends by setting the speed to `-xBargeDriveSpeed` or `xBargeDriveSpeed` so there's a smooth transition. Might be better to just combine `driveToBlackLine` and `driveToBarge` into a single composite command that manages that transition & safely stops the robot if it's interrupted.

jamesdooley4 (dismissed):

src/main/java/frc/robot/Controls.java
KangarooKoala: > This seems odd, as `atScoringXPosition` being true will interrupt `driveToBarge` and should also then cancel `algaeNetScore` since it's all part of the whole command composition group.

The `.until()` only interrupts the command it decorates. If we were using a race command (or the driving command was the deadline), then the drive command ending (whether interrupted or normally) would then interrupt the scoring command, but since we aren't, the driving command can finish early without affecting the scoring command.

> Though maybe `withDeadline` overrides the `until()` interrupt and extends the command group until `algaeNetScore` ends even though `driveToBarge().until()` would have already ended?

Yep! Deadline commands (like [`cmdA.withDeadline(cmdB)`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Command.html#withDeadline(edu.wpi.first.wpilibj2.command.Command))) end when and only when the deadline ends. The deadline will keep on running if the other commands finish (like what happens in this case), and if other commands are still running when the deadline finishes, they will be interrupted.

src/main/java/frc/robot/subsystems/auto/BargeAlign.java
KangarooKoala: Copy-pasta from the controls drive request, which in turn is from the original generated CTRE source. We should definitely remove those stale comments.

src/main/java/frc/robot/Controls.java
KangarooKoala: Good point about the abrupt stop!

I'm a bit iffy about doing the sequence logic ourselves inside the class command instead of via a sequential, but It'd probably be manageable. We would need to sort out how to cause the scoring sequence to start at the right time, though.

jamesdooley4: Maybe pass in a command to run when the black line is reached to the BargeAlign command?

KangarooKoala: We could make that work, but it'd make the logic harder to reason about. (Coordinating sequences and parallels of actions can't really be cleanly expressed by logic in `execute()`, and manually scheduling commands is a code smell.)

The cleanest solution is probably making a `Command BargeAlign.bargeAlign(CommandSwerveDrivetrain, SuperStructure)` method that encapsulates the logic of driving to the black line, then moving to the black line and scoring. That way, we can safely make `end()` not stop the drive base, since we can make the separate steps private and guarantee the drive to black line command will always be followed by the drive to barge command.

src/main/java/frc/robot/Controls.java
```diff
@@ -349,7 +350,16 @@ private void configureSuperStructureBindings() {
...
+                                        .until(
+                                            () ->
+                                                BargeAlign.atScoringXPosition(s.drivebaseSubsystem))
```
KangarooKoala, line 357: We should put the `.until()` inside the definition of `driveToBarge()`.

Jetblackdragon (approved):
Looks good from 4/8 testing

PR #135, opened 2025-04-07
Remove elevator soft limits

RobototesProgrammers:
Apparently leader motor hitting soft limit doesn't stop the follower


KangarooKoala (approved):
Tested 4/7.

PR #137, opened 2025-04-07
Ground intake improvements

### LINKED ISSUE #131: Tune ground arm movement
We should find how high we can raise the motion magic cruise velocity, acceleration, and jerk. We may also want to lower P (since we were only raising it in an attempt to make it go faster before we realized we should be increasing the motion magic settings instead), but it might be fine as it is.

RobototesProgrammers:
Also fix typo in 3 piece path name, use the right competition MAC address, and make arm pivot faster.

Closes #131.


KangarooKoala (approved):
Tested 4/7.

PR #142, opened 2025-04-08
Autos post dcmp

Jetblackdragon:


KangarooKoala (approved):
Tested 4/8. 3 piece is a 2.8 piece if we get an AprilTag reading as we approach C, otherwise it's a 2 piece.

PR #150, opened 2025-06-04
Elastic layout implementation

RobototesProgrammers:


src/main/java/frc/robot/Robot.java
```diff
@@ -107,6 +108,7 @@ protected Robot() {
       FollowPathCommand.warmupCommand().schedule();
     }
     MatchTab.create(sensors, subsystems);
+    WebServer.start(5800, Filesystem.getDeployDirectory().getPath());
```
markpete, line 111: Is this needed even in competition?  Could potentially slow other operations, so perhaps debug only?

markpete: Given that it's broadly agreed this is needed, and the use case explanation, I'll close this issue.

markpete (approved):

PR #156, opened 2025-08-16
Fixes for various issues

### LINKED ISSUE #88: More LED indicators
From discussion with drive team, we should add the following indicators:
* [ ] ~~Ready to score (arm and elevator at right position, aligned with branch): Flash green at 3 Hz~~ We just score when branch sensors aligned, no need to tell drive team.
* [x] Elevator position is less than -1: LEDs solid red
* [x] 30 seconds before match end: Briefly blink red and yellow
* [x] 15 seconds before match end: Briefly blink red and yellow

iamawesomecat:
Current fixes:

Added red LED flashing when the elevator is giving incorrect position readings while Zero'd, should help find faulty Zero's as they happen. (resolves #88 )

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -669,7 +669,7 @@ public class Controls {
                   .blink(120, 0, 0, "Red - Elevator Not Zeroed")
                   .ignoringDisable(true));
       subZeroPosition.whileTrue(
-          Commands.waitSeconds(0.2)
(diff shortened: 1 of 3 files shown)


TrisDooley (dismissed):
L1 scoring broken
https://youtube.com/shorts/ECKsj3ImBwg?si=zJd4w0h4opQcDULT

Jetblackdragon (approved):

Issue #157, opened 2025-08-17
Improve auto human loading consistency for the 3 piece auto


TrisDooley:
Current situation:
The robot does not wait long enough at the human player station to be loaded consistently, and will raise without a coral in the claw after leaving.

Proposed fixes:
- Raise the amount of time the robot is waiting at the station, and have this OR with the coral sensing.
- Require that the ground intake -> claw handoff has occurred and the claw senses a coral before raising the elevator. If the handoff occurs and no coral is sensed, jiggle a bit and retry.

TrisDooley:
Fixed through 158

PR #158, opened 2025-08-17
Humanplayerautochanges

Jetblackdragon:
#157 added a check to make sure coral can be in took and increased timeout for coral at station

### DIFF
--- a/src/main/deploy/pathplanner/autos/OSMRSF_E-D.auto
+++ b/src/main/deploy/pathplanner/autos/OSMRSF_E-D.auto
@@ -36,12 +36,6 @@
           }
         },
         {
-          "type": "path",
-          "data": {
-            "pathName": "E to RSF"
-          }
-        },
-        {
           "type": "named",
           "data": {
             "name": "isCollected"
--- a/src/main/java/frc/robot/subsystems/SuperStructure.java
+++ b/src/main/java/frc/robot/subsystems/SuperStructure.java
@@ -1,5 +1,9 @@
 package frc.robot.subsystems;
 
+import java.util.Set;
+import java.util.function.BooleanSupplier;
+import java.util.function.Supplier;
+
 import edu.wpi.first.wpilibj2.command.Command;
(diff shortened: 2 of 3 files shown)


Jetblackdragon:
New working (probably) in claw logic. This can be tested if you disable vision and put the robot on a dolly so the wheels spin but it doesnt move and try feeding it vertically when it goes to the human player station (can check this using advantage scopes readings).

TrisDooley (changes requested):
The robot still tries to score without holding a coral 
https://youtube.com/shorts/Ny060lzi6RQ?si=P5JJS2oTzbLcrzjj

Jetblackdragon:
So I have no idea why the elevator is moving even though it knows it's in claw is false. I will keep looking into it. The in intake uses the lasercan in the ground intake and Im guessing on of the limits is just a tiny bit off which should be a quick fix.

TrisDooley (commented):
So very close to working! I think this is good enough to prevent catastrophic coral crushing, but it's still raising the mini funnel which allows coral under it.
https://youtube.com/shorts/AQCaoq1911M?si=YmRPuNsJerMIYlQP

Jetblackdragon:
Changed it so that in autos doesn't raise as high and will prevent it from going under.

TrisDooley:
This seems to work? Normal autos no longer score more than 1 piece though.

No coral, only 1 score: https://youtube.com/shorts/-yzhZ9_9VKU?si=v14hhsqK1Q2mN4Ua

No starting, but then fed coral 0 score: https://youtube.com/shorts/39DnoICOEMs?si=TqYoXJ1jiC6PW4pT

Starting and fed coral, 1 score: https://youtube.com/shorts/eWtfIPQgoxc?si=XCF4-tYnvYFaleMA

Jetblackdragon:
Should work now with some of the fixes that werent mergeed earlier.

TrisDooley (commented):
Only one coral is scored: https://youtube.com/shorts/4G_ZxkJRIhA?si=6R-cLxcEDQzhMPLT

Issue #164, opened 2025-08-28
Funnel stow position inaccuracies.


Jetblackdragon:
When going into coral mode the funnel ends up in different places and sometimes gets coral stuck.

PR #169, opened 2025-09-09
More advanced climb controller

TrisDooley:
Starting work on #164


TrisDooley:
Undid the screw-up from drive practice

iamawesomecat:
I approve merging this (as climb subsystem lead), but its Aarya's call.

iamawesomecat (approved):
Aarya is ok with it 👍

PR #175, opened 2025-09-25
Auto algae elevator

Jetblackdragon:
Solo Controls + Elevator auto moves to the right algae intake height

### DIFF
--- a/src/main/java/frc/robot/subsystems/SuperStructure.java
+++ b/src/main/java/frc/robot/subsystems/SuperStructure.java
@@ -312,6 +312,19 @@ public class SuperStructure {
         .withName("Coral Stow");
   }
 
+  public Command autoCoralStow() {
+    return Commands.defer(
+            () ->
+                Commands.parallel(
+                    elevator.setLevel(ElevatorSubsystem.CORAL_LEVEL_TWO_PRE_POS),
+                    Commands.sequence(
+                        Commands.waitUntil(elevator.above(ElevatorSubsystem.CORAL_PRE_INTAKE)),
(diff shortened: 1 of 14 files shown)


src/main/java/frc/robot/subsystems/SuperStructure.java
```diff
@@ -393,6 +393,20 @@ public Command algaeLevelTwoThreeIntake() { // removes algae from lower reef
         .withName("Algae Ground Intake");
   }
 
+  public Command algaeLevelThreeFourIntakeWithoutElevator() { // removes algae from upper reef
```
markpete, line 396: These two commands seem identical except for the logging.  Why are two different commands needed here?  If you need different logging, make it a single command, but pass it a parameter to define what "level" it's at and use that for logging.

src/main/java/frc/robot/subsystems/auto/AutoAlgaeElevatorCommand.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  private static boolean isBlue() {
+    return DriverStation.getAlliance()
+        .map(alliance -> alliance.equals(Alliance.Blue))
+        .orElse(false);
+  }
+
+  private static Pose2d getNearestAlgae(Pose2d robotPose) {
+    List<Pose2d> poses = isBlue() ? blueAlgaePoses : redAlgaePoses;
```
markpete, line 56: We do this a bunch in different functions.  We should add this to the class instantiation and do it once instead of for every time a function is called.

Issue #177, opened 2025-09-29
Rework algae sequence


iamawesomecat:
With our current setup, Algae is too inconsistent to reliably do during a match.

PR #179, opened 2025-09-29
Making barge align use vision instead of odometry

koolpoolo:
helps address #177

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -30,6 +30,7 @@ import frc.robot.subsystems.ArmPivot;
 import frc.robot.subsystems.ElevatorSubsystem;
 import frc.robot.subsystems.GroundArm;
 import frc.robot.subsystems.SuperStructure;
+import frc.robot.subsystems.auto.AutoAlgaeHeights;
 import frc.robot.subsystems.auto.AutoAlign;
 import frc.robot.subsystems.auto.BargeAlign;
 import frc.robot.util.AlgaeIntakeHeight;
@@ -60,7 +61,7 @@ public class Controls {
 
   private BranchHeight branchHeight = BranchHeight.CORAL_LEVEL_FOUR;
   private ScoringMode scoringMode = ScoringMode.CORAL;
-  private ScoringMode intakeMode = ScoringMode.CORAL;
+  public ScoringMode intakeMode = ScoringMode.CORAL;
   private SoloScoringMode soloScoringMode = SoloScoringMode.NO_GAME_PIECE;
   private AlgaeIntakeHeight algaeIntakeHeight = AlgaeIntakeHeight.ALGAE_LEVEL_THREE_FOUR;
 
@@ -735,7 +736,7 @@ public class Controls {
     }
 
     s.elevatorLEDSubsystem.setDefaultCommand(
-        s.elevatorLEDSubsystem.showScoringMode(() -> soloScoringMode));
+        s.elevatorLEDSubsystem.showScoringMode(() -> soloScoringMode, intakeMode));
 
     if (s.elevatorSubsystem != null) {
       Trigger hasBeenZeroed = new Trigger(s.elevatorSubsystem::getHasBeenZeroed);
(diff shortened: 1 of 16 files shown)


src/main/java/frc/robot/Controls.java
```diff
@@ -128,7 +129,8 @@ private double getDriveY() {
     return input * MaxSpeed * inputScale;
   }
 
-  // takes the rotation value from the joystic, and applies a deadband and input scaling
+  // takes the rotation value from the joystic, and applies a deadband and input
```
markpete, line 132: typo:  joystic -> joystick

src/main/java/frc/robot/Controls.java
```diff
@@ -161,30 +163,31 @@ private void configureDrivebaseBindings() {
                         .withRotationalRate(getDriveRotate()))
             .withName("Drive"));
 
-    // various former controls that were previously used and could be referenced in the future
+    // various former controls that were previously used and could be referenced in
+    // the future
 
     // operatorController
```
markpete, line 169: How likely is this to be used in the future?  If it hasn't been used in 6 months, just delete it.

src/main/java/frc/robot/subsystems/auto/BargeAlign.java
```diff
@@ -27,17 +28,19 @@ public class BargeAlign extends Command {
           .withDriveRequestType(DriveRequestType.OpenLoopVoltage)
           .withForwardPerspective(ForwardPerspectiveValue.BlueAlliance);
 
-  public static boolean atScoringXPosition(CommandSwerveDrivetrain drivebasesubsystem) {
-    double robotX = drivebasesubsystem.getState().Pose.getX();
+  public static boolean atScoringXPosition(DrivebaseWrapper drivebaseWrapper) {
+    double robotX = drivebaseWrapper.getEstimatedPosition().getX();
```
markpete, line 32: Can drivebaseWrapper be null?  Always good to do a null check before dereferencing a variable.

markpete:
Passing driveBaseWrapper around to all of the routines is fine and aligns with the programming standards the team uses, so fine leave as is.  However, assuming there is ever only a single driveBaseWrapper instance at a time, you might consider implementing a Singleton pattern inside of the DriveBaseWrapper class, that way you can get the instance inside of the routines where you need it instead of having to pass it down through the call stacks all over the place.

Note that this is a bit more of an advanced coding construct (so totally fine not to do it this way).
https://www.geeksforgeeks.org/system-design/singleton-design-pattern/

markpete (changes requested):
Please address comments.

src/main/java/frc/robot/Controls.java
```diff
@@ -60,7 +60,8 @@ public class Controls {
   private AlgaeIntakeHeight algaeIntakeHeight = AlgaeIntakeHeight.ALGAE_LEVEL_THREE_FOUR;
 
   // Swerve stuff
-  // setting the max speed nad other similar variables depending on which drivebase it is
+  // setting the max speed nad other similar variables depending on which
```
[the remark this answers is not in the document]
koolpoolo: done

src/main/java/frc/robot/Controls.java
koolpoolo: done

src/main/java/frc/robot/Controls.java
koolpoolo: no its for archives

src/main/java/frc/robot/Controls.java
```diff
@@ -597,12 +601,14 @@ private void configureClimbPivotBindings() {
     s.climbPivotSubsystem.setDefaultCommand(
         s.climbPivotSubsystem.advanceClimbCheck().withName("Advance Climb Check"));
 
-    // check if the climb controller is connected, and whne start is pressed move to the next climb
+    // check if the climb controller is connected, and whne start is pressed move to
```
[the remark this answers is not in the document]
koolpoolo: done

PR #183, opened 2025-10-05
All the offseason changes put together.

Jetblackdragon:


src/main/java/frc/robot/Controls.java
```diff
@@ -151,7 +151,8 @@ private void configureDrivebaseBindings() {
       // Stop running this method
       return;
     }
-
+    // if the solo controller is connected, use it for driving
+    boolean soloCC = soloController.isConnected();
```
jamesdooley4, line 155: I think this will result in `soloCC` only being set when the robot code starts running, so changing to or from soloCC will require a restart of the robot code. Consider instead making this a boolean class member variable and then updating it ever second or so based on whether the solo controller is connected (via adding a new `update` method in this class that `Robot.robotPeriodic()` calls, and have the `update` method do the soloCC update every second or so).

RobototesProgrammers: it worked fine at drive practice we can check again at the meeting on monday

PR #199, opened 2025-11-20
Choreo routines

SnappleRamen:
This branch added Choreo autos that the robot can run in simulation.


src/main/java/frc/robot/subsystems/drivebase/CommandSwerveDrivetrain.java
```diff
@@ -178,6 +187,28 @@ public Command applyRequest(Supplier<SwerveRequest> requestSupplier) {
     return run(() -> this.setControl(requestSupplier.get())).withName("Drivebase applyRequest()");
   }
 
+  public void followTrajectory(SwerveSample sample) {
+    // Get the current pose of the robot
+    Pose2d pose = this.getState().Pose;
+    // Pose2d branchPose = AutoAlignCommand.getNearestBranch(pose);
```
markpete: Better to remove lines that you no longer need rather than comment them out.

markpete:
There appear to be a bunch of declarations that are made as public and I don't really see the reason you did that.  If they are not needed outside of the class, make them private.

markpete (changes requested):

PR #200, opened 2025-11-20
AutoBuilder now builds upon selection rather than auto init

SnappleRamen:
Optimization improvements to the AutoBuilder. The Autos would be preloaded and ready as soon as a new Auto is selected, rather than having to do all that at the start of Auto.

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -144,8 +144,8 @@ public class Robot extends TimedRobot {
   @Override
   public void autonomousInit() {
     Shuffleboard.startRecording();
-    if (SubsystemConstants.DRIVEBASE_ENABLED && AutoLogic.getSelectedAuto() != null) {
-      AutoLogic.getSelectedAuto().schedule();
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/Robot.java
```diff
@@ -144,8 +145,10 @@ public void disabledExit() {
   @Override
   public void autonomousInit() {
     Shuffleboard.startRecording();
-    if (SubsystemConstants.DRIVEBASE_ENABLED && AutoLogic.getSelectedAuto() != null) {
-      AutoLogic.getSelectedAuto().schedule();
+    if (SubsystemConstants.DRIVEBASE_ENABLED) {
+      Commands.waitSeconds(AutoLogic.autoDelayEntry.get().getDouble())
```
jamesdooley4: While you're here, it'd be nice to skip the waitSeconds call if the autoDelayEntry entry is 0. As written (and as currently implemented), I think this delays the auto routine by a code cycle since the WaitCommand has to run for a single cycle to detect that it has nothing to do.

PR #208, opened 2025-11-25
fix #197

### LINKED ISSUE #197: coralLevelThree is nearly identical to coralLevelTwo
`coralLevelThree()` is identical to `coralLevelTwo()`, just with different constants for a few things. If they should match other than constants, it'd be useful to rework this post-comp to remove the duplicated code (via a template function or some other method)

_Originally posted by @jamesdooley4 in https://github.com/robototes/Reefscape2025/pull/189#discussion_r2462593846_

SnappyRiffs:
fixed issue #197-combined level two and level three into one function

### DIFF
--- a/src/main/java/frc/robot/subsystems/SuperStructure.java
+++ b/src/main/java/frc/robot/subsystems/SuperStructure.java
@@ -185,41 +185,39 @@ public class SuperStructure {
   // if robot is in position for intermediate scoring, it can score early but if vision is dead
   // manual control still works
   // only used on solo controls
+  public Command coralLevelTwo(BooleanSupplier score, BooleanSupplier ipScore) { // same as L4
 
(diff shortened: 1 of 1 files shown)


src/main/java/frc/robot/subsystems/SuperStructure.java
markpete: See if you can remove redundancy on this function too, by having it call the new routine.

HINT:  ipScore = () -> { return true };

markpete: This routine should be able to leverage the new shared code you wrote as well.

markpete: You should also be able to combine the two coralLevelOne Commands.

markpete (changes requested):
There is still more opportunity for removing redundant code.

HINT:  ipScore = () -> { return true; }

PR #211, opened 2025-11-26
Modernize Gradle configuration

markpete:
Modernize Gradle configuration by introducing version catalogs, setting build properties, and updating Java toolchain.
 
 This should simplify config management and speed up build times.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,6 @@
+# Enable build caching
+org.gradle.caching=true
+# Enable parallel execution
+org.gradle.parallel=true
+# Configure JVM memory
+org.gradle.jvmargs=-Xmx2g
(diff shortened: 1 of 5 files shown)


koolpoolo (approved):
looks good to me as it centralizes the versions as well which is nice to have them in one neat place

gradle.properties
```diff
@@ -0,0 +1,6 @@
+# Enable build caching
+org.gradle.caching=true
+# Enable parallel execution
+org.gradle.parallel=true
+# Configure JVM memory
```
iamawesomecat: Will any of these configurations interfere with how the roborio runs and deploys code? I dont have much understanding on gradle, just want to make sure that it wont lead to any unintentional side affects with roborio performance.

iamawesomecat: (aka, can we test this during todays meeting? @markpete @Jetblackdragon)
I know this may be troublesome to our schedule with the prototyping challenge, so I understand if theres any issues there.

jamesdooley4 (approved):

PR #217, opened 2025-12-04
Gains tuning utility class

iamawesomecat:
A network tables based class to assist in PID tuning. Can be dropped into any subsystem and fully utilized in like 10 lines of code. Use case showed in armPivot changes.

Has a constructor for:

P,I,D config (Basic PID)
P,I,D,K,S,V,G config (certain control types for Talon FX motors do not allow for kA, so I made a constructor w/o it)
P,I,D,K,S,A,V,G config (every feedback & feedforward constant)

### DIFF
--- a/src/main/java/frc/robot/subsystems/ArmPivot.java
+++ b/src/main/java/frc/robot/subsystems/ArmPivot.java
@@ -4,6 +4,7 @@ package frc.robot.subsystems;
 import static edu.wpi.first.units.Units.Seconds;
 import static edu.wpi.first.units.Units.Volts;
 
+import com.ctre.phoenix6.configs.Slot0Configs;
 import com.ctre.phoenix6.configs.TalonFXConfiguration;
 import com.ctre.phoenix6.configs.TalonFXConfigurator;
 import com.ctre.phoenix6.controls.MotionMagicVoltage;
@@ -28,6 +29,7 @@ import edu.wpi.first.wpilibj2.command.button.Trigger;
 import edu.wpi.first.wpilibj2.command.sysid.SysIdRoutine;
 import edu.wpi.first.wpilibj2.command.sysid.SysIdRoutine.Direction;
 import frc.robot.Hardware;
+import frc.robot.util.GainsTuningEntry;
 import java.util.function.Supplier;
 
 public class ArmPivot extends SubsystemBase {
@@ -74,6 +76,18 @@ public class ArmPivot extends SubsystemBase {
 
   // TalonFX
   private final TalonFX motor;
+
+  // Testing tuning entry :P
+  private final GainsTuningEntry PIDtuner =
+      new GainsTuningEntry(
+          "ArmPivot",
+          ARMPIVOT_KP,
+          ARMPIVOT_KI,
+          ARMPIVOT_KD,
+          ARMPIVOT_KA,
+          ARMPIVOT_KV,
+          ARMPIVOT_KS,
+          ARMPIVOT_KG);
 
   // alerts if motor is not connected.
   private final Alert NotConnectedError =
@@ -227,6 +241,16 @@ public class ArmPivot extends SubsystemBase {
   public void periodic() {
     // Error that ensures the motor is connected
     NotConnectedError.set(notConnectedDebouncer.calculate(!motor.getMotorVoltage().hasUpdated()));
+
+    Slot0Configs pidTunerConfigs = new Slot0Configs();
+    pidTunerConfigs.kP = PIDtuner.getP();
+    pidTunerConfigs.kI = PIDtuner.getI();
+    pidTunerConfigs.kD = PIDtuner.getD();
+    pidTunerConfigs.kS = PIDtuner.getS();
+    pidTunerConfigs.kV = PIDtuner.getV();
+    pidTunerConfigs.kA = PIDtuner.getA();
+    pidTunerConfigs.kG = ARMPIVOT_KG;
+    motor.getConfigurator().apply(pidTunerConfigs);
   }
 }
 //  -Samuel "Big North" Mallick
--- a/src/main/java/frc/robot/subsystems/ArmPivot.java
+++ b/src/main/java/frc/robot/subsystems/ArmPivot.java
@@ -34,13 +34,13 @@ import java.util.function.Supplier;
 
 public class ArmPivot extends SubsystemBase {
   // Presets
-  private final double ARMPIVOT_KP = 38.5; // previously 50
-  private final double ARMPIVOT_KI = 0;
-  private final double ARMPIVOT_KD = 0;
-  private final double ARMPIVOT_KS = 0.1;
-  private final double ARMPIVOT_KV = 0.69;
-  private final double ARMPIVOT_KG = 0.18;
-  private final double ARMPIVOT_KA = 0.0;
+  private double armPivot_kP = 38.5; // previously 50
+  private double armPivot_kI = 0;
+  private double armPivot_kD = 0;
+  private double armPivot_kS = 0.1;
+  private double armPivot_kV = 0.69;
+  private double armPivot_kG = 0.18;
+  private double armPivot_kA = 0.0;
   // Preset positions for Arm with Coral
   public static final double CORAL_PRESET_L1 = 0;
   public static final double CORAL_PRESET_L2 = 0.13;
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/subsystems/ArmPivot.java
```diff
@@ -227,6 +241,16 @@ public void factoryDefaults() {
...
+
+    Slot0Configs pidTunerConfigs = new Slot0Configs();
+    pidTunerConfigs.kP = PIDtuner.getP();
+    pidTunerConfigs.kI = PIDtuner.getI();
+    pidTunerConfigs.kD = PIDtuner.getD();
+    pidTunerConfigs.kS = PIDtuner.getS();
+    pidTunerConfigs.kV = PIDtuner.getV();
+    pidTunerConfigs.kA = PIDtuner.getA();
+    pidTunerConfigs.kG = ARMPIVOT_KG;
+    motor.getConfigurator().apply(pidTunerConfigs);
```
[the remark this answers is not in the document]
iamawesomecat: the pidTuner code in ArmPivot should be disabled when tuning is not ongoing.

src/main/java/frc/robot/util/GainsTuningEntry.java
```diff
@@ -0,0 +1,120 @@
...
+    return kDEntry.getDouble(0);
+  }
+
+  public double getA() {
+    if (kAEntry == null) {
+      return -1;
+    } else {
+      return kAEntry.getDouble(0);
+    }
+  }
```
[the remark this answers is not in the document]
iamawesomecat: fixed, idk why gemini thinks I didnt

src/main/java/frc/robot/util/GainsTuningEntry.java
```diff
@@ -0,0 +1,120 @@
...
+  private NetworkTableEntry kAEntry;
+  private NetworkTableEntry kVEntry;
+  private NetworkTableEntry kSEntry;
+  private NetworkTableEntry kGEntry;
+
+  /* Base PID Feedback entires. Excludes kA, kV, kS, and kG
+   *
+   */
+  public GainsTuningEntry(String SubsystemName, double kp, double ki, double kd) {
+    table = NetworkTableInstance.getDefault().getTable("Gains/" + SubsystemName);
```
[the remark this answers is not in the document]
iamawesomecat: fixed

src/main/java/frc/robot/subsystems/ArmPivot.java
```diff
@@ -75,6 +77,18 @@ public class ArmPivot extends SubsystemBase {
   // TalonFX
   private final TalonFX motor;
 
+  // Testing tuning entry :P
```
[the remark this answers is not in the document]
iamawesomecat: let me be joyous

src/main/java/frc/robot/subsystems/ArmPivot.java
```diff
@@ -227,6 +242,17 @@ public void factoryDefaults() {
...
+    if (TUNING_ENABLED) {
+      Slot0Configs pidTunerConfigs = new Slot0Configs();
+      pidTunerConfigs.kP = PIDtuner.getP();
+      pidTunerConfigs.kI = PIDtuner.getI();
+      pidTunerConfigs.kD = PIDtuner.getD();
+      pidTunerConfigs.kS = PIDtuner.getS();
+      pidTunerConfigs.kV = PIDtuner.getV();
+      pidTunerConfigs.kA = PIDtuner.getA();
+      pidTunerConfigs.kG = PIDtuner.getG();
+      motor.getConfigurator().apply(pidTunerConfigs);
```
jamesdooley4, line 273: Per [applying configs](https://v6.docs.ctr-electronics.com/en/latest/docs/api-reference/api-usage/configuration.html#applying-configs) this blocks for an while to wait for the config to apply. Since it's unlikely the config is changing 50x a second, it should only apply a new config if it's changed compared the previously applied config.

jamesdooley4 (changes requested):

src/main/java/frc/robot/util/GainsTuningEntry.java
```diff
@@ -0,0 +1,176 @@
...
+        || currentG != getG());
+  }
+
+  /* Check if any gains have changed using the current values from the subsystem.
+   * Compares P, I, D only
+   */
+  public boolean pidGainsChanged(double currentP, double currentI, double currentD) {
+
+    return (currentP != getP() || currentI != getI() || currentD != getD());
+  }
```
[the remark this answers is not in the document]
iamawesomecat: fixed

src/main/java/frc/robot/util/GainsTuningEntry.java
```diff
@@ -0,0 +1,150 @@
...
+    kVEntry.setDouble(kv);
+    kSEntry.setDouble(ks);
+    kGEntry.setDouble(kg);
+  }
+
+  public String getSubsystem() {
+    return kSubsystemName;
+  }
+
+  /** Get proportional gain (P) */
```
Jetblackdragon, line 72: All of these could be simplified into one method that you just pass in the value you are looking for. Unnecessary but cuts down the repetitiveness.

src/main/java/frc/robot/subsystems/ArmPivot.java
```diff
@@ -227,6 +241,37 @@ public void factoryDefaults() {
...
+        armPivot_kV,
+        armPivot_kS,
+        armPivot_kG)) {
+      armPivot_kP = PIDtuner.getP();
+      armPivot_kI = PIDtuner.getI();
+      armPivot_kD = PIDtuner.getD();
+      armPivot_kS = PIDtuner.getS();
+      armPivot_kV = PIDtuner.getV();
+      armPivot_kA = PIDtuner.getA();
+      armPivot_kG = PIDtuner.getG();
```
Jetblackdragon, line 260: Instead of doing this periodically you could add a button that updates your values. For example you add button in elastic that says update and it pushes the new values instead of constantly checking and updating the value in 2 places. Its also safety incase you reconfigure it while its running a command (Im not sure what the priority on that is tho). Also unnecessary but something to consider.

Jetblackdragon (commented):
Looks good, I suggested some small things but its not something I am too worried about.

src/main/java/frc/robot/util/GainsTuningEntry.java
Jetblackdragon: the lines for this is wrong its from line 73 to something

src/main/java/frc/robot/subsystems/ArmPivot.java
markpete, line 239: Needs a better comment header

src/main/java/frc/robot/subsystems/ArmPivot.java
```diff
@@ -227,6 +241,37 @@ public void factoryDefaults() {
...
+      pidTunerConfigs.kP = armPivot_kP;
+      pidTunerConfigs.kI = armPivot_kI;
+      pidTunerConfigs.kD = armPivot_kD;
+      pidTunerConfigs.kS = armPivot_kS;
+      pidTunerConfigs.kV = armPivot_kV;
+      pidTunerConfigs.kA = armPivot_kA;
+      pidTunerConfigs.kG = armPivot_kG;
+
+      // applying the new gains to the motor
+      motor.getConfigurator().apply(pidTunerConfigs);
```
markpete, line 273: It might be better to just store the configurator object locally, so that you don't have to retrieve it every time through this loop.  Because this loop is periodic, it can happen often, so you want to optimize for fast run-time.

```
if (!!m_Configurator) {
   m_Configurator = motor.getConfigurator();
}
m_Configurator.apply(pidTunerConfigs);
```

src/main/java/frc/robot/util/GainsTuningEntry.java
```diff
@@ -0,0 +1,150 @@
...
+  private NetworkTableEntry kDEntry;
+  private NetworkTableEntry kAEntry;
+  private NetworkTableEntry kVEntry;
+  private NetworkTableEntry kSEntry;
+  private NetworkTableEntry kGEntry;
+
+  /** Base PID Feedback entires. Excludes kA, kV, kS, and kG */
+  public GainsTuningEntry(String subsystemName, double kp, double ki, double kd) {
+    table = NetworkTableInstance.getDefault().getTable("Gains/" + subsystemName);
+    this.kSubsystemName = subsystemName;
```
markpete, line 22: Is there any validation you need to do on any of these parameters?  e.g. null or empty subsystemName, negative or null values?  What assumptions are you making about the data that need to be confirmed?

markpete (commented):
A few suggestions...  Would be great to incorporate, but otherwise looks good.
I do agree with @Jetblackdragon 's suggestion to consider tying this to a button to trigger rather than making it automatic.  If you get a bad reading you're hosed until the next periodic update.  (E.g. it slips on the carpet, collisions,  or something...)

Either way, you'll want to test this especially how bad readings might impact.

